PR #31, opened 2016-07-04
compaction_metrics

nirmaayan:
Add two graphs 

Compaction Summary:

```
{
  "aliasColors": {},
  "bars": false,
  "datasource": "prometheus",
  "editable": true,
  "error": false,
  "fill": 1,
  "grid": {
    "threshold1": null,
    "threshold1Color": "rgba(216, 200, 27, 0.27)",
    "threshold2": null,
    "threshold2Color": "rgba(234, 112, 112, 0.22)"
  },
  "id": 42,
  "isNew": true,
  "legend": {
    "avg": false,
    "current": false,
    "max": false,
    "min": false,
    "show": false,
    "total": false,
    "values": false
  },
  "lines": true,
  "linewidth": 2,
  "links": [],
  "nullPointMode": "connected",
  "percentage": false,
  "pointradius": 5,
  "points": false,
  "renderer": "flot",
  "seriesOverrides": [],
  "span": 3,
  "stack": false,
  "steppedLine": false,
  "targets": [
    {
      "expr": "sum(collectd_compaction_manager_objects) by (instance)",
      "intervalFactor": 2,
      "refId": "A",
      "step": 10,
      "metric": "collectd_compaction_manager_objects",
      "hide": false
    }
  ],
  "timeFrom": null,
  "timeShift": null,
  "title": "Compaction Summary",
  "tooltip": {
    "msResolution": false,
    "shared": true,
    "value_type": "cumulative"
  },
  "type": "graph",
  "xaxis": {
    "show": true
  },
  "yaxes": [
    {
      "format": "short",
      "logBase": 1,
      "max": null,
      "min": 0,
      "show": true
    },
    {
      "format": "short",
      "logBase": 1,
      "max": null,
      "min": null,
      "show": true
    }
  ],
  "transparent": false
}
```

Compaction AVG:

```
{
  "aliasColors": {},
  "bars": false,
  "datasource": "prometheus",
  "editable": true,
  "error": false,
  "fill": 1,
  "grid": {
    "threshold1": null,
    "threshold1Color": "rgba(216, 200, 27, 0.27)",
    "threshold2": null,
    "threshold2Color": "rgba(234, 112, 112, 0.22)"
  },
  "id": 43,
  "isNew": true,
  "legend": {
    "avg": false,
    "current": false,
    "max": false,
    "min": false,
    "show": false,
    "total": false,
    "values": false
  },
  "lines": true,
  "linewidth": 2,
  "links": [],
  "nullPointMode": "connected",
[...]

PR #32, opened 2016-07-05
format Grafana json file using json_reformat

tzach:
This PR reformat the JSON file using linux `json_reformat` command.
From now own, all updates to JSON files in this project will keep this formatting style.


tzach:
@nirmaayan please pull this update, and use `json_reformat` for #31

PR #41, opened 2016-08-25
use docker inspect to get prometheus ip

tzach:
This request bypass hacks used to provide the grafana instance the IP of the prometheus image, by using docker inspect. This makes the special EC2 and local use case redundant, and one can always use start-all.sh

A new script kill-all.sh, kill and remove the two docker instances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,18 +21,18 @@ ubuntu $ sudo systemctl restart docker
 centos $ sudo service docker start
 ```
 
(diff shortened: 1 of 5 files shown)


tzach:
@penberg @nirmaayan can you please review?

tzach:
image

penberg:
@tzach Interesting. Will that map help with `docker inspect`? :wink:

tzach:
> @tzach Interesting. Will that map help with docker inspect? 

oops, wrong issue

PR #46, opened 2016-08-31
add metrics about the I/O Queue

glommer:
There are way too many I/O Queue metrics, so let's move them to their own
dashboard. Usually one will be looking at just a couple of metrics at once, but
that's because only some classes usually play an important role in the load.

Since the reason we have too many metrics is that we have a lot of classes,
there really isn't a way to simplify this for the main dashboard unless we leave
out really important information.

For each of the classes, 3 out of the 4 metrics are exposed:
- delay, meaning for how long requests sit in the queue
- bandwidth, self-explanatory
- IOPS, self-explanatory. In conjunction with bandwidth can be
  used to find out the average size of requests.

Signed-off-by: Glauber Costa glommer@scylladb.com

### DIFF
--- a/grafana/scylla-dash-io-per-server.json
+++ b/grafana/scylla-dash-io-per-server.json
@@ -1,0 +1,2292 @@
+{
+  "id": 5,
+  "title": "Scylla Per-Server Disk I/O",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "hideControls": true,
+  "sharedCrosshair": true,
+  "rows": [
+    {
+      "collapse": false,
+      "editable": true,
+      "height": "25px",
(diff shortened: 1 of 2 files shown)


tzach:
@glommer the new API failed to work (return 422)
I suspect the issue is the json file, which should start with 

```
{
    "dashboard": {
        "id": null,

```

Check out the existing dashboard as an example
https://github.com/scylladb/scylla-grafana-monitoring/blob/master/grafana/scylla-dash.json

glommer:
Thanks, I will look.

But I was able to upload and use this dashboard just fine.

nirmaayan:
I took the json format and create the dashboard manually and it worked perfectly

tzach:
closing this as v2 was already merged.

PR #47, opened 2016-09-01
add metrics about the I/O Queue

glommer:
There are way too many I/O Queue metrics, so let's move them to their own
dashboard. Usually one will be looking at just a couple of metrics at once, but
that's because only some classes usually play an important role in the load.

Since the reason we have too many metrics is that we have a lot of classes,
there really isn't a way to simplify this for the main dashboard unless we leave
out really important information.

For each of the classes, 3 out of the 4 metrics are exposed:
- delay, meaning for how long requests sit in the queue
- bandwidth, self-explanatory
- IOPS, self-explanatory. In conjunction with bandwidth can be
  used to find out the average size of requests.

Signed-off-by: Glauber Costa glommer@scylladb.com

### DIFF
--- a/grafana/scylla-dash-io-per-server.json
+++ b/grafana/scylla-dash-io-per-server.json
@@ -1,0 +1,2294 @@
+{
+    "dashboard": {
+      "id": null,
+      "title": "Scylla Per-Server Disk I/O",
+      "tags": [],
+      "style": "dark",
+      "timezone": "browser",
+      "editable": true,
+      "hideControls": true,
+      "sharedCrosshair": true,
+      "rows": [
+        {
+          "collapse": false,
+          "editable": true,
+          "height": "25px",
+          "panels": [
+            {
+              "content": "<img src=\"http://www.scylladb.com/img/logo.svg\" height=\"70\">\n<hr style=\"border-top: 3px solid #5780c1;\">",
+              "editable": true,
+              "error": false,
+              "id": 41,
+              "isNew": true,
+              "links": [],
+              "mode": "html",
+              "span": 12,
+              "title": "",
+              "transparent": true,
+              "type": "text"
+            }
+          ],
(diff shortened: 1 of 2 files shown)


tzach:
Screen capture
image

glommer:
side note, I feel a bit disappointed that grafana will show powers of 10 of seconds, instead of microseconds, milliseconds, etc.

tzach:
One way to do it to multipal by 1000, and change the unit on the Y Axis
image
image

glommer:
That's not a good way to do it. This will just shift the problem from "seconds" to "milliseconds" (what if we are taking nanosecond?)

What I would like to do is see grafana auto adjusting this. It does it for bandwidth (shows 10^9 bps as 1Gbps), but not for time.

glommer:
@tgrabiec just posted a panel to Scylla issue 1634 where the graph adapts to the right time unit.

Tomek, how did you do it? At least in my grafana installation, keeping the I/O queue delay in "seconds" show ugly powers of 10 that don't scale well.

tgrabiec:
@glommer the axis ticks switch automatically to larger units (but not smaller). For example, if the value is in [ns], 1000 ns will be shown on the axis as 1 [us]. If the metric is inherently in [ms] but can take fractional values, then you can multiply it to convert to [ns].

glommer:
Wow, that is very counterintuitive.

I will change them to microseconds then

tzach:
> Wow, that is very counterintuitive.

Indeed, maybe we should open an issue for the Grafana project

glommer:
On Fri, Sep 2, 2016 at 9:43 AM, Tzach Livyatan notifications@github.com
wrote:

> Wow, that is very counterintuitive.
> 
> Indeed, maybe we should open an issue for the Grafana project

https://github.com/grafana/grafana/issues/5955

PR #54, opened 2016-09-08
add Thrift operation metrics

tzach:
This request adds Thrift to operation charts

### DIFF
--- a/grafana/scylla-dash-io-per-server.json
+++ b/grafana/scylla-dash-io-per-server.json
@@ -1,2294 +1,2495 @@
 {
     "dashboard": {
-      "id": null,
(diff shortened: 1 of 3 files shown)


tzach:
@glommer please review

tzach:
@tgrabiec ?

tgrabiec:
Me and Glauber think that it would be better to sum the CQL and thrift requests on the "Request served per Server" and "Total requests served", etc, so that there is a single line per server. And maybe have separate graphs somewhere else for CQL and thrift servers, but that could come later.

tzach:
Thanks @glommer @tgrabiec 
I will send a v2

PR #59, opened 2016-09-19
prometheus: Set honor_labels to true

tgrabiec:
I have a setup with a single collectd sink instance (ccm cluster) and
therefore single collectd_exporter. In such configuration all samples
will be assigned by prometheus the same instance label - the collectd
exporter's IP. They do however have distinct instance label set by the
exporter. Setting honor_labels to true makes prometheus respect the
label, which fixes the per-server dashboard in such setups.

### DIFF
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -8,5 +8,6 @@ global:
 
 scrape_configs:
 - job_name: scylla
+  honor_labels: true
   static_configs:
   - targets: ['127.0.0.1.18:9103']


tgrabiec:
I can see that this change was already merged in #57

tzach:
@tgrabiec yes, sorry for not notify you on time

tzach:
@tgrabiec the problem with this fix, is the nodes are no longer can be identified by ip, rather by some obscure code. This is much less useful. Any advice?

tgrabiec:
@tzach What is "some obscure code"? They should be identified by what was passed to `--collectd-hostname`.

tzach:
> They should be identified by what was passed to --collectd-hostname.

@tgrabiec is this a parameter for collectd? if so, I did not set it, which explain the random string I get

tgrabiec:
@tzach This is seastar's parameter. If you don't specify it `gethostname()` name will be used.

tzach:
Thanks @tgrabiec 
My mistake was running Scylla docker image without `-h` for hostname. Docker uses a random string in this case, which propagates to my Grafana dashboard.
@penberg this is something to note when you adding monitoring into the official Scylla Docker image

I updated my [Scylla in a box](https://gist.github.com/tzach/1975e1f0d7964e10783979a174745755) to include hostname.

PR #61, opened 2016-09-25
upload dashboards dir instead of using REST API

### LINKED ISSUE #48: Mount Grafana dashboard directory
Current solution uploads dashboard using REST API. This is error prone and requires opening the REST API which might be a security issue.
A better solution will be to mount Grafana dashboard directory, as done with prometheus yaml file.

tzach:
fix #48


tzach:
The result of this update is the host grafana/dashboards dir owner changed to qemu after starting the grafana docker. Not sure if this is a problem or not.
@tgrabiec thoughts?

tzach:
@nirmaayan @kongove can you please test this PR?
It attaches the dashboard directory to Grafana container, rather than uploading it using REST API.
In particular, try to update the dashboard, save it, use it later etc
Thanks

PR #64, opened 2016-09-26
Fix configure typo and update doc of loading history data

amoskong:
The 1st patch fixed a typo of ip address in prometheus.yml

The 2nd patch updated README about loading history data to prometheus server. User may report a issue with a prometheus data, we also download prometheus data in SCT performance testing, it's a easy way to review the history data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,25 @@ Update `prometheus/prometheus.yml` with the targets (server you wish to monitor)
 ./start-all.sh
 ```
 
+### Load original data to prometheus server
+
+```
+Additional parameters:
+  -v <data>:/prometheus
+
+Full commandline:
+  sudo docker run -d -v <data>:/prometheus -v $PWD/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml:Z -p 9090:9090 --name aprom prom/prometheus:v1.0.0
+
+Comment:
+  <data> is the local path to original data directory
+
+Data source:
+1. Download from docker prometheus server, reference:
+https://github.com/scylladb/scylla/wiki/How-to-report-a-Scylla-problem#prometheus
+2. Get from Scylla-Cluster-Test log.
+3. etc
+```
+
 ### Kill
 
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,6 +29,19 @@ Update `prometheus/prometheus.yml` with the targets (server you wish to monitor)
 ./start-all.sh
 ```
 
+### Load original data to prometheus server
```
tzach, line 32: @kongove can you explain where this data comes from?
Is it the same data capture in this procedure:
https://github.com/scylladb/scylla/wiki/How-to-report-a-Scylla-problem#prometheus

prometheus/prometheus.yml
```diff
@@ -10,4 +10,4 @@ scrape_configs:
 - job_name: scylla
   honor_labels: true
   static_configs:
-  - targets: ['127.0.0.1.18:9103']
+  - targets: ['127.0.0.1:9090']
```
tzach: why 9090? port 9103 is the default  collectd_exporter port, as describe here
https://github.com/scylladb/scylla/wiki/Monitor-Scylla-with-Prometheus-and-Grafana#collectd_exporter

tzach:
@kongove thanks for the PR, it make a lot of sense
Two comments above

README.md
amoskong: Yes, it's same. I loaded the prometheus data which got from SCT performance test, grafana can present it correctly. It's helpful.

Data source:
1. Download from docker prometheus server, reference:
https://github.com/scylladb/scylla/wiki/How-to-report-a-Scylla-problem#prometheus
2. Get from Scylla-Cluster-Test log.
3. etc

prometheus/prometheus.yml
amoskong: I referenced the example in prometheus github [1]. I will keep using 9103 port when I rebase the PR. 

[1] https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus.yml

README.md
tzach: Thanks, please added this info to the README itself, so users will make the connection, and can go back to Scylla wiki. Its not obvious to everyone

PR #67, opened 2016-09-29
add more cache and lsa graphs

pdziepak:
Added graphs:
- cache insertions
- cache evictions
- cache partitions
- cache total bytes
- lsa memory:
  - lsa total memory
  - unused memory in lsa segments
  - non-lsa used memory
  - free memory

### DIFF
--- a/grafana/scylla-dash-per-server.json
+++ b/grafana/scylla-dash-per-server.json
@@ -1522,7 +1522,7 @@
                         },
                         "type": "graph",
                         "xaxis": {
-                            "show": false
+                            "show": true
                         },
                         "yaxes": [
                             {
@@ -1602,7 +1602,7 @@
                         },
                         "type": "graph",
                         "xaxis": {
-                            "show": false
+                            "show": true
                         },
                         "yaxes": [
                             {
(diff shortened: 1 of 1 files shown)


tzach:
@pdziepak thanks for the PR
The data chart looks good and useful

The new chart is positioned under Disk, which is not correct.
image

I suggest dedicating a new row for it, with a new **Memory** header in a new row (see below)
image

I suspect the new chart have too many data lines and will be unreadable with many nodes, but it can be fixed in a later patch. Maybe we will have a dashboard just for memory.

pdziepak:
Ah, I didn't realise that in more than a multiple node scenarios we would have data from all nodes on these graphs. LSA memory graph is stacked so it would make absolutely no sense at all in such cases.

For tracking LSA memory the best would be a separate stacked graph for each node, I am not sure whether it can be done with grafana. Now, I think it is better to drop it and leave just the new cache graphs.

tzach:
> For tracking LSA memory the best would be a separate stacked graph for each node, I am not sure whether it can be done with grafana. 

I think we should use http://docs.grafana.org/reference/templating/  but I haven't tried it yet

tgrabiec:
@pdziepak  In per-server view we could have separate charts with just LSA (per-node), non-LSA (per-node), etc.  It would be still useful, I think.

pdziepak:
I've split the LSA memory graph to "LSA total memory" and "non-LSA memory" and moved it to a separate section.

screenshot from 2016-09-29 17-21-38

tzach:
Looks good
image

PR #81, opened 2016-11-06
use max instead of sum when dealing with time delays

glommer:
We use sum() to aggregate IOPS and bandwidth, but are using it by mistake
to aggregate delays. That makes little sense, and for time we should be using
max() instead.


tzach:
@glommer I assume #85 make this PR redundant
Can we close it?

glommer:
Yes

PR #86, opened 2016-11-25
grafana: per-server: Use step 1

### LINKED ISSUE #45: Grafana graphs are jiggly with short scraping intervals
When I set scraping interval to 1s I see graphs "jiggle" in a way that historical data points change values. After looking closer at this it turns out that graphs alternate on refreshes between showing samples from only odd and only even timestamp values.

Setting "resolution" parameter for queries from `1/2` to `1/1` fixes the problem. In dashboard JSON the change amounts to:

``` patch
           "targets": [
             {
               "expr": "avg(collectd_reactor_gauge{type=\"load\"} ) by (instance)",
-              "intervalFactor": 2,
+              "intervalFactor": 1,
               "refId": "A",
-              "step": 2
+              "step": 1
             }
           ],
           "timeFrom": null,
```

tgrabiec:
When I set scraping interval to 1s I see graphs "jiggle" in a way that
historical data points change values. After looking closer at this it
turns out that graphs alternate on refreshes between showing samples
from only odd and only even timestamp values.

Setting "resolution" parameter for queries from 1/2 to 1/1 fixes the
problem.

Fixes #45

### DIFF
--- a/grafana/scylla-dash-per-server.json
+++ b/grafana/scylla-dash-per-server.json
@@ -82,10 +82,10 @@
                         "targets": [
                             {
                                 "expr": "count(up{job=\"scylla\"})",
-                                "intervalFactor": 2,
+                                "intervalFactor": 1,
                                 "legendFormat": "",
                                 "refId": "A",
-                                "step": 40
+                                "step": 1
                             }
                         ],
                         "thresholds": "",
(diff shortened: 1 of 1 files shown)


amnonh:
@tgrabiec there is currently a conflict with the commit, also, is the problem still relevant after the move to the Prometheus API?

tgrabiec:
@amnonh Yes, the problem exists with connecting to Scylla over the prometheus API.

tzach:
Thanks @tgrabiec 

* I still see a case of` "step": 4 ` in grafana/scylla-dash-per-server.json (line 382)
and  `"intervalFactor": 2` (line 380) 

* Shouldn't we make the same change in all 3 dashboards?

* I have a Conflicting files issue (probably a later merge)

tgrabiec:
@tzach Problems 1 and 3 are due to concurrent merges, will rebase. Will change other dashboards as well.

PR #94, opened 2016-11-28
grafana: Use intervalFactor 1

tgrabiec:
grafana: Use resolution of 1/1

When I set scraping interval to 1s I see graphs "jiggle" in a way that
historical data points change values. After looking closer at this it
turns out that graphs alternate on refreshes between showing samples
from only odd and only even timestamp values. This is because of
intervalFactor being set to 1/2, even when auto-tuned step is small.

Setting "resolution" parameter for queries from 1/2 to 1/1.

The problem may still happen when the step is larger than scrape
period, which can happen when viewing a long time range. There the
solution would be to set the step to 1s. But I am not sure we want to
do this by default, as it would make grafana unusuable when viewing
long time ranges due to amount of data which would need to be fetched.

Refs #45

### DIFF
--- a/grafana/scylla-dash-per-server.json
+++ b/grafana/scylla-dash-per-server.json
@@ -82,10 +82,10 @@
                         "targets": [
                             {
                                 "expr": "count(up{job=\"scylla\"})",
-                                "intervalFactor": 2,
+                                "intervalFactor": 1,
                                 "legendFormat": "",
                                 "refId": "A",
-                                "step": 40
+                                "step": 1
                             }
                         ],
                         "thresholds": "",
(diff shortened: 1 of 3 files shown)


grafana/scylla-dash-per-server.json
```diff
@@ -377,7 +377,7 @@
                         "targets": [
                             {
                                 "expr": "avg(seastar_reactor{type=\"gauge\", metric=\"load\"} ) by (instance)",
-                                "intervalFactor": 2,
+                                "intervalFactor": 1,
                                 "refId": "A",
                                 "step": 4
```
tzach, line 382: @tgrabiec you left the step here as 4.
Is it intentional?

tgrabiec: Yes it's intentional. The step doesn't seem to be  relevant in this change. I don't actually  know what it signifies.

PR #95, opened 2016-11-29
Add graphs with replica-side counters

tgrabiec:
Right below the storage proxy section. Preview:

new_graphs

### DIFF
--- a/grafana/scylla-dash-per-server.json
+++ b/grafana/scylla-dash-per-server.json
@@ -65,6 +65,17 @@
                         "links": [
 
                         ],
+                        "mappingType": 1,
+                        "mappingTypes": [
+                            {
+                                "name": "value to text",
+                                "value": 1
+                            },
+                            {
+                                "name": "range to text",
+                                "value": 2
+                            }
+                        ],
                         "maxDataPoints": 100,
                         "nullPointMode": "connected",
                         "nullText": null,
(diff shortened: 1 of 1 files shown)


tzach:
@tgrabiec looks good, but I do not sure I understand the metrics.
What does Replica mean here, and what each of the new dashboard stands for?
Thanks

tgrabiec:
@tzach In Scylla, requests execute in distributed manner, possibly on many nodes. The node to which the client sends a query is called "coordinator" node (or storage_proxy internally). The coordinator then contacts "replica" nodes which hold relevant data. An update for instance will issue write requests to RF number of replicas. A query may read from multiple replica nodes in order to ensure CL. 

The "Replica" section introduced here groups metrics for replica role of the node(s).

It shows:
  - "Reads" - rate of reads (rps) served by this replica (currently only completed, due to lack of counters, will extend later)
  - "Writes" - rate of writes (wps) served by this replica (currently only completed, due to lack of counters, will extend later)
  - "Writes currently blocked on dirty" - number of active writes which are blocked due to limits of dirty (unflushed) memory. It's memory used by memtables.
  - "Writes blocked on dirty" - rate of writes blocked due to above
  - "Writes currently blocked on commitlog" - number of active writes which are currently blocked due limit on memory which can be in unflushed commitlog buffers. Blocking here means commitlog doesn't keep up.
  - "Writes blocked on commitlog" - rate of writes blocked due to above
  - "Active sstable reads" - number of currently active sstable reads (there is at most one such read per replica read, if it missed the cache)
  - "Queued sstable reads" - there is a limit on active sstable reads, when exceeded, reads queue, and this counter shows the length of that queue.

tzach:
Thanks @tgrabiec 
I'm merging this since it looks useful already.
As a second step, we will need to:
- include some of your descriptions on the dashboard itself
- create a user-centric dashboard. The per server dashboard has become expert land.  Terms like "Writes blocked on dirty"  are hard to understand (and sound like a Roger Corman movie ;)

PR #97, opened 2016-11-30
grafana: replica: Add failure counters

tgrabiec:
New graphs with rates of :
  - failed reads
  - failed writes
  - timedout writes

failure_counters

### DIFF
--- a/grafana/scylla-dash-per-server.json
+++ b/grafana/scylla-dash-per-server.json
@@ -1256,10 +1256,9 @@
                 "title": "New row"
             },
             {
-                "title": "New row",
-                "height": "250px",
-                "editable": true,
                 "collapse": false,
+                "editable": true,
+                "height": "250px",
                 "panels": [
                     {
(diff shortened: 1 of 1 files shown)


tzach:
@tgrabiec I do not see any data when looking at total_reads_failed, same for writes.
Are these new metrics in Scylla?

tgrabiec:
@tzach They were merged to master yesterday.

tzach:
> They were merged to master yesterday.

Aha. So there are no users who have metrics yet, but it will be immediately visible and confusing for anyone who uses the latest monitoring stack (as we recommend)

Should we start branching this project with correlating branches to Scylla, or even move the dashboards under Scylla?
@penberg @avikivity thoughts?

tgrabiec:
@tzach Same is true already for when you look at 1.5 metrics while being connected to 1.4 Scylla. It's not a problem introduced by this pull.

tzach:
Indeed, its a problem we need to solve, not just this PR
Even Scylla 1.4 users are using the latest master Grafana dashboard

tzach:
Merging an empty chart for older Scylla version is a minor issue.
We will need to put order into monitoring releases.

PR #100, opened 2016-12-06
Update example prometheus.yml

tzach:
This PR improve the usability of prometheus.yml:

* update the examples to include two servers with relevant ports
* consistently use ```'``` rather than a mix of ```"``` and ```'```


tzach:
@amnonh please review

PR #103, opened 2016-12-22
prometheus: Reduce scrape period to 1 second

### LINKED ISSUE #93: Use 1 second scrape period by default
Current period of 15 seconds will average out variations in utilization which happen with period shorter than 15 seconds. Using shorter interval will give us more information about system's behavior. 1s should be still large enough to not put significant load on the system.

tgrabiec:
Current period of 15 seconds will average out variations in
utilization which happen with period shorter than 15 seconds. Using
shorter interval will give us more information about system's
behavior. 1s should be still large enough to not put significant load
on the system.

Fixes #93.

### DIFF
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -1,5 +1,5 @@
 global:
-  scrape_interval: 15s # By default, scrape targets every 15 seconds.
+  scrape_interval: 1s # By default, scrape targets every 1 second.
 
   # Attach these labels to any time series or alerts when communicating with
   # external systems (federation, remote storage, Alertmanager).


tzach:
Thanks @tgrabiec 

The default is 15s, and other projects are using 5s or 10s for Cassandra monitoring[1]
I'm afraid 1s will collect too much info and fill the disk quickly, or will be too often for larger clusters.

@avikivity @amnonh  Thoughts?
We want the default to fit users needs.


[1] https://github.com/outbrain/Cassibility/blob/master/prometheus.yml

tgrabiec:
15s is too coarse grained for spotting many anomalies. I've been asking users which report performance issues to reduce scrape period to 1s. Having that as a default would save us the trouble of explaining this, we would get higher accuracy telemetry by default.

There's clearly a trade-off here. You can run out of disk space with 15s scrape period as well. It will let you cover longer time range, but with lower precision.

tzach:
OK, if it's more useful for us to find issue with, let's use 1s as default.

PR #118, opened 2017-02-15
bring back proxy mode

### LINKED ISSUE #115: initialized: undefined failure, Default prometheus container address wrong?
Hi,
A fresh installation of a monitoring solution fails to start.
The error message received on the screen is: 
image

Looking into the data sources, the ip address setting for the prometheus database is set to 127.0.0.1:9090 and direct

Looking into another working monitoring installation, it is 172.17.0.2:9090 and proxy based.

Going back to the freshly installed monitoring and modifying the IP of the datasource to 172.17.0.2:9090 and to proxy access, it solves the issue.

Options for resolution:
1. We document the needed change on the DB IP addresses
2. Ship the monitoring solution with an IP that is defined in the installation of DB

glommer:
Local mode, albeit much simpler, doesn't work in remote installations.
In those, the browser will probe for prometheus directly and at
localhost it will be nowhere to be found.

We can still have multi-port mode working in proxy mode by making sure
we probe the correct container and noticing that we always have to
connect to its port 9090, and not the locally bound port.

Fixes #115

### DIFF
--- a/start-all.sh
+++ b/start-all.sh
@@ -113,8 +113,12 @@ then
         exit 1
 fi
 
+# Can't use localhost here, because the monitoring may be running remotely.
+# Also note that the port to which we need to connect is 9090, regardless of which port we bind to at localhost.
+DB_ADDRESS="$(sudo docker inspect --format '{{ .NetworkSettings.IPAddress }}' $PROMETHEUS_NAME):9090"
+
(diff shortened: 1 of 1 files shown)


tzach:
Looks good to me.
@eyalgutkind can you please validate it solve the problem for you?

tzach:
Issue solved on EC2 so I'm merging.
@eyalgutkind please open a new issue if the problem persists.

Thanks @glommer

PR #119, opened 2017-02-16
update dashboards refresh to 30s

### LINKED ISSUE #114: 5 second refresh time is to fast
On a slower network, it can cause bad experience. 30 s should be enough.

tzach:
fix #114


tzach:
@amnonh please review

PR #126, opened 2017-02-17
add option to specify nodes

glommer:
Editing prometheus.yml is a bit of a pain, and now more so that we have
to add each node twice. Add an option in start-all.sh to do it
automatically given nodes passed in the command line.

### DIFF
--- a/gen-prometheus.py
+++ b/gen-prometheus.py
@@ -1,0 +1,27 @@
+template = """
+global:
+  scrape_interval: 1s # By default, scrape targets every 1 second.
+
+  # Attach these labels to any time series or alerts when communicating with
+  # external systems (federation, remote storage, Alertmanager).
+  external_labels:
+    monitor: 'scylla-monitor'
+
+scrape_configs:
+- job_name: scylla
+  honor_labels: true
+  static_configs:
+  - targets: %s
+
+- job_name: node_exporter
+  honor_labels: true
+  static_configs:
+  - targets: %s
+"""
+
+from sys import argv
+from string import split
+ips=split(argv[1],",")
+def add_port(vec, port):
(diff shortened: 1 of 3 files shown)


tzach:
@glommer Great idea, which can be made simpler with extracting the server list to an extrnal file[1]
prometheus.yml will include something like:

```yaml
scrape_configs:
- job_name: scylla
  honor_labels: true
  file_sd_configs:
  - files:
      - scylla_servers.yaml
```
and scylla_servers.yaml will look like:

```yaml
- targets: ['127.0.0.2:9180']
```
(the same for node_exporter)

This will make the templating redundent, and can allow adding target to the file on runtime with `refresh_interval`

Thoughts?

[1] https://prometheus.io/docs/operating/configuration/#file_sd_config

glommer:
I was not aware if that option.

I think it makes sense.


On Feb 19, 2017 3:26 AM, "Tzach Livyatan" <notifications@github.com> wrote:

@glommer <https://github.com/glommer> Great idea, which can be made simpler
with extracting the server list to an extrnal file[1]
prometheus.yml will include something like:

scrape_configs:
- job_name: scylla
  honor_labels: true
  file_sd_configs:
  - files:
      - scylla_servers.yaml

and scylla_servers.yaml will look like:

- targets: ['127.0.0.2:9180']

(the same for node_exporter)

This will make the templating redundent, and can allow adding target to the
file on runtime with refresh_interval

Thoughts?

[1] https://prometheus.io/docs/operating/configuration/#file_sd_config

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/scylladb/scylla-grafana-monitoring/pull/126#issuecomment-280904072>,
or mute the thread

PR #132, opened 2017-02-20
move prometheus targets from prometheus.yml to two new yml files

tzach:
The new targets files are easier to generate dynamically from external scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,19 +30,24 @@ ubuntu $ sudo systemctl restart docker
 centos $ sudo service docker start
 ```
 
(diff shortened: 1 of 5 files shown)


tzach:
@glommer @amn please review

glommer:
Tested and it works fine.

amn:
@tzach I am not the contributor you were looking for.

tzach:
@amn you can be if you want to :) sorry for calling you
@amnonh I was referring to you

PR #136, opened 2017-02-21
add option to specify nodes

glommer:
Editing the configuration files manually  is a bit of a pain, and now
more so that we have to add each node twice. Add an option in
start-all.sh to do it automatically given nodes passed in the command
line.

### DIFF
--- a/gen-prometheus.py
+++ b/gen-prometheus.py
@@ -1,0 +1,7 @@
+from sys import argv
+from string import split
+ips=split(argv[1],",")
+port=argv[2]
+print "- targets:"
+for addr in map(lambda x: x + ":" + port, ips):
+    print "  - " + addr
--- a/kill-all.sh
+++ b/kill-all.sh
@@ -32,7 +32,10 @@ if [ -z $PROMETHEUS_PORT ]; then
 else
     PROMETHEUS_NAME=aprom-$PROMETHEUS_PORT
 fi
-
+SCYLLA_NODES_FILE="$PWD/prometheus/scylla-$PROMETHEUS_PORT.yml"
+NODE_EXPORTER_NODES_FILE="$PWD/prometheus/node-$PROMETHEUS_PORT.yml"
 
 sudo docker kill $GRAFANA_NAME $PROMETHEUS_NAME
 sudo docker rm $GRAFANA_NAME $PROMETHEUS_NAME
+rm $SCYLLA_NODES_FILE 2>/dev/null
+rm $NODE_EXPORTER_NODES_FILE 2>/dev/null
--- a/start-all.sh
+++ b/start-all.sh
@@ -2,12 +2,12 @@
 
 . versions.sh
 VERSIONS=$DEFAULT_VERSION
-usage="$(basename "$0") [-h] [-d Prometheus data-dir] [-v comma seperated versions] [-g grafana port ] [ -p prometheus port ] -- starts Grafana and Prometheus Docker instances"
+usage="$(basename "$0") [-h] [-d Prometheus data-dir] [-v comma seperated versions] [-g grafana port ] [ -p prometheus port ] [-n comma separated list of nodes to monitor ] -- starts Grafana and Prometheus Docker instances"
 
 GRAFANA_VERSION=4.1.1
 PROMETHEUS_VERSION=v1.5.2
 
-while getopts ':hd:g:p:v:' option; do
+while getopts ':hd:g:p:v:n:' option; do
   case "$option" in
     h) echo "$usage"
        exit
@@ -19,6 +19,8 @@ while getopts ':hd:g:p:v:' option; do
     g) GRAFANA_PORT=$OPTARG
        ;;
     p) PROMETHEUS_PORT=$OPTARG
+       ;;
+    n) NODES=$OPTARG
        ;;
     :) printf "missing argument for -%s\n" "$OPTARG" >&2
        echo "$usage" >&2
@@ -44,6 +46,16 @@ else
     PROMETHEUS_NAME=aprom-$PROMETHEUS_PORT
 fi
 
+if [ -z $NODES ]; then
+    NODES="127.0.0.1"
+fi
+# Don't put in a temporary location. The file needs to be still present if we are
+# to restart the container upon reboot (for example)
+SCYLLA_NODES_FILE="$PWD/prometheus/scylla_servers-$PROMETHEUS_PORT.yml"
+NODE_EXPORTER_NODES_FILE="$PWD/prometheus/node_exporter_servers-$PROMETHEUS_PORT.yml"
+python gen-prometheus.py $NODES 9180 > $SCYLLA_NODES_FILE
+python gen-prometheus.py $NODES 9100 > $NODE_EXPORTER_NODES_FILE
+
 # Exit if Docker engine is not running
 if [ ! "$(sudo docker ps)" ]
 then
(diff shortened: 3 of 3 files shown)


tzach:
@amnonh can you please review?

gen-prometheus.py
```diff
@@ -0,0 +1,7 @@
+from sys import argv
+from string import split
+ips=split(argv[1],",")
+port=argv[2]
+print "- targets:"
+for addr in map(lambda x: x + ":" + port, ips):
```
amnonh, line 6: if you are using python, not why use a yamel library?

glommer:
On Tue, Feb 21, 2017 at 11:33 AM, Amnon Heiman <notifications@github.com>
wrote:

> *@amnonh* commented on this pull request.
> ------------------------------
>
> In gen-prometheus.py
> <https://github.com/scylladb/scylla-grafana-monitoring/pull/136#discussion_r102252928>
> :
>
> > @@ -0,0 +1,7 @@
> +from sys import argv
> +from string import split
> +ips=split(argv[1],",")
> +port=argv[2]
> +print "- targets:"
> +for addr in map(lambda x: x + ":" + port, ips):
>
> if you are using python, not why use a yamel library?
>

I'm not that much of an expert - so I don't know any. Not to mention this
is such a simple iterator, and 90 %  of the code is iterating through argv
and preparing the strings.

I doubt any library could make it simpler.

amnonh:
It's not must, but something like:

http://stackoverflow.com/questions/12470665/how-can-i-write-data-in-yaml-format-in-a-file
this way you'll have a library that read and write yaml, in case you'll want to add stuff to it in the future

glommer:
On Tue, Feb 21, 2017 at 11:59 AM, Amnon Heiman <notifications@github.com>
wrote:

> It's not must, but something like:
>
> http://stackoverflow.com/questions/12470665/how-can-i-
> write-data-in-yaml-format-in-a-file
> this way you'll have a library that read and write yaml, in case you'll
> want to add stuff to it in the future
>

Thanks for the pointer - but it does sound like overkill for the moment,
and this file is not likely to change much.
Do you agree with that ?

kill-all.sh
```diff
@@ -32,7 +32,10 @@ if [ -z $PROMETHEUS_PORT ]; then
 else
     PROMETHEUS_NAME=aprom-$PROMETHEUS_PORT
 fi
-
+SCYLLA_NODES_FILE="$PWD/prometheus/scylla-$PROMETHEUS_PORT.yml"
+NODE_EXPORTER_NODES_FILE="$PWD/prometheus/node-$PROMETHEUS_PORT.yml"
 
 sudo docker kill $GRAFANA_NAME $PROMETHEUS_NAME
 sudo docker rm $GRAFANA_NAME $PROMETHEUS_NAME
+rm $SCYLLA_NODES_FILE 2>/dev/null
```
amnonh, line 40: you probably wants to add -f just to be sure that rm will not fail

glommer:
On Tue, Feb 21, 2017 at 12:25 PM, Amnon Heiman <notifications@github.com>
wrote:

> *@amnonh* commented on this pull request.
> ------------------------------
>
> In kill-all.sh
> <https://github.com/scylladb/scylla-grafana-monitoring/pull/136#discussion_r102266797>
> :
>
> >
>  sudo docker kill $GRAFANA_NAME $PROMETHEUS_NAME
>  sudo docker rm $GRAFANA_NAME $PROMETHEUS_NAME
> +rm $SCYLLA_NODES_FILE 2>/dev/null
>
> you probably wants to add -f just to be sure that rm will not fail
>

I can do that.

start-all.sh
```diff
@@ -44,22 +46,40 @@ else
     PROMETHEUS_NAME=aprom-$PROMETHEUS_PORT
 fi
 
+if [ -z $NODES ]; then
+    NODES="127.0.0.1"
```
amnonh, line 50: Does that mean that you can not use a local file? if no nodes are supplied just take the configuration file

gen-prometheus.py
amnonh: You need a python header comment (either python or python3)
For what you're doing here, you can just add bash loop (like the one that handle versions in start-all.sh)

If you prefer to keep the python, just add the header and I'll merge it

kill-all.sh
glommer: Ok, will do that.

start-all.sh
glommer: I personally feel like that having two deployment methods is confusing (local conf file and -nodes switch). That's why I standardized everything on nodes, and defaulted to localhost if no node is passed.

For us it is great to be as flexible as possible but I think it may confuse users. @tzach what do you think ?

tzach:
I have to say I'm ambiguous about this PR
It looks like much of extra complexity (python, adding and removing files) which brings value only for a few users. I'm afraid this extra complexity will break the project in unexpected ways.
I understand it's useful for us; I wonder if we can do it in a less intrusive/painful way. Maybe there is none.

@glommer @amnonh thoughts?

glommer:
@tzach Allow me to disagree (although I am obviously biased).

Everybody that uses scylla-grafana-monitoring needs to update the node list. Everybody.

Once you have hundreds of nodes, it's way simpler to do that by passing a list of nodes than editing files.

tzach:
> Once you have hundreds of nodes, it's way simpler to do that by passing a list of nodes than editing files.

Passing hundreds of IPs as command line parameter? not sure it's a good alternative as well.
For larger systems, DevOps are using tools like [Consul](https://www.consul.io/) to hold the cluster IPs. Prometheus can read the targets from Consul.

I agree provisioning hundreds of targets is not easy with text files as well.
The question is, does this PR bring value to a production user with 3-10 nodes cluster?
Wouldn't it be easier for him to set the target list once, and not on every restart?

glommer:
For a small deployment monitoring a single cluster it indeed doesn't make much sense - although it is marginally simpler to use the command line if you are updating the repository: since prometheus.yaml is tracked by git, it will always be reset on updates.

This provides values in my view for people running large clusters, or for people like me, monitoring many clusters from the same machine.

I am obviously fine with alterntive solutions - not married to this one. But please note that if we're not taking neither, I'd suggest we revert the previous patch that splits the file: I believe changing the addresses in two files is more error-prone than one. Easy to forget.

tzach:
> For a small deployment monitoring a single cluster it indeed doesn't make much sense - although it is marginally simpler to use the command line if you are updating the repository: since prometheus.yaml is tracked by git, it will always be reset on updates.

True. This is super annoying

>  I am obviously fine with alternative solutions - not married to this one. But please note that if we're not taking neither, I'd suggest we revert the previous patch that splits the file: I believe changing the addresses in two files is more error-prone than one. Easy to forget.

Might be. Let me sleep on it and take an extra day to talk with users. Sorry for slowing you down on this one.

glommer:
don't worry @tzach. Doing it right is more important than doing it fast.

tzach:
We need to keep an option to have seperate target list for scylla and node_exporter.
In a Docker base system, we will have more of the first and less of the second.

tzach:
@amnonh @glommer a few mental iterations I came to the following requirements:
- most users will use file (or files) of targets, which rarely change
- we need a way to specify different targets for node_exporter and scylla (altho its annoying)

The following will improve UX:
- a way to provide a command line option for either the target list or better a file with the target list
- a way to keep the target list file(s) out of the git

So one can run:
```
./start-all.sh --scylla-targets scylla-straget.yaml --node-targets node-straget.yaml 
```
or even better using the same target yaml file for both (not sure its possible due to different port)

This will allow users to keep their target files, and run different instances with different target lists.
BTW, the end goal of the monitoring system is monitoring multiple clusters in the same system, with a cluster/DC selection. Running multiple instances of Prometheus is not the right way (but we can allow it, as suggested)

Thoughts?

tzach:
Closing this as #138 gives similar functionality

PR #138, opened 2017-02-23
allow using non default target files

tzach:
This PR allows one to use his target files like

```
./start-all.sh -s my_scylla_servers.yaml -n my_node_exporter_servers.yml -d data_dir
```

In combination with the ports option `-g` and `-p` this can be used to run multiple Grafana/Prometheus stacks, each with different targets

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ For example, update targets in `prometheus/node_exporter_servers.yaml` :
   - 172.17.0.3:9100
 ```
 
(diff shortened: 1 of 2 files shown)


tzach:
@amnonh @glommer please review this alternative to #136

tzach:
@amnonh @glommer ping

glommer:
I will look at the code now. In terms of functionality, it works for me and I think it is generic enough -> so that's good.

glommer:
Looks good to me.

tzach:
Thanks @glommer

PR #156, opened 2017-04-02
Remove a container and its volume

### LINKED ISSUE #155: kill-all.sh skript leaves volumes behind
after couple of management restarts disk may run out of space.

tzach:
Fix #155

### DIFF
--- a/kill-all.sh
+++ b/kill-all.sh
@@ -35,4 +35,4 @@ fi
 
 
 sudo docker kill $GRAFANA_NAME $PROMETHEUS_NAME
-sudo docker rm $GRAFANA_NAME $PROMETHEUS_NAME
+sudo docker rm -v $GRAFANA_NAME $PROMETHEUS_NAME


tzach:
@gleb-cloudius please test.
The question is, does the user ALWAYS wants to remove the volumes?
@glommer @amnonh thoughts ?

glommer:
I think so. After the kill script is invoked the containers are gone, and the only way to get to them is through the filesystem. Without inspecting the container you can't even know where in the filesystem it is.

Add an option to leave the directories intact and if we really care, but default to delete

gleb-cloudius:
On Sun, Apr 02, 2017 at 06:17:19AM -0700, Tzach Livyatan wrote:
> @gleb-cloudius please test.
We already tested it with Amnon.

--
			Gleb.

tzach:
Thanks for the reviewers! So far its the smallest PR I sent or merged. Lets see who can do better.

PR #162, opened 2017-04-29
split grafana loading phase into its own script

glommer:
There are deployments in which, due to firewall rules or other issues,
once has issues pulling docker images. Or alternatively, an user may
already have a prometheus/grafana installation that they may want to
reuse.

For situations like that, it will be useful to extract the container
setup phase from the data loading phase. Users can then call just the
latter.

### DIFF
--- a/load-grafana.sh
+++ b/load-grafana.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+. versions.sh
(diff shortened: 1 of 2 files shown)


tzach:
Looks good
@glommer can you send PR for the readme, and/or update the Wiki with instructions?
Users will not guess it there otherwise.

PR #165, opened 2017-05-03
fix cache metrics graphs

glommer:
Graphs showing currently blank, since metrics had their names changed in
1.7. Fix them, so they'll start showing.


glommer:
It seems to me I was using a master AMI by accident. Sorry

PR #167, opened 2017-05-03
script to generate yaml configuration files

glommer:
We currently have to edit the yml files by hand to add the nodes ip
addresses to the files. This is hard to automate, and for this reason
I am bringing this script:

It will generate scylla_servers.yml and node_exporter.yml files into
a directory of the user's choosing. It can write both files or one of
them only, which we can control by the -s and -n flags respectively.

usage: genconfig.py [-h] [-d DIRECTORY] [-s] [-n] node_ip [node_ip ...]

Generate configuration for prometheus

positional arguments:
  node_ip               list of nodes to configure, separated by space

optional arguments:
  -h, --help            show this help message and exit
  -d DIRECTORY, --directory DIRECTORY
                        directory where to generate the configuration files
  -s, --scylla          Generate scylla_servers.yml file
  -n, --node            Generate node_exporter_servers.yml file

### DIFF
--- a/genconfig.py
+++ b/genconfig.py
@@ -1,0 +1,39 @@
+#!/usr/bin/python
+
+import argparse
+import os
+import yaml
+
+scylla_port=9180
+node_exporter_port=9100
+
+def append_port(ips, port):
+    return [ "%s:%s"%(x, port) for x in ips ]
+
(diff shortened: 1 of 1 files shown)


tzach:
@glommer little by little, you are building consul

Issue #124, opened 2017-02-16
Provide Prometheus retention value


tzach:
Without a retention value, Prometheus will fill up the disk.
Prometheus expose this value with `-storage.local.retention` parameter.
The actual value depends on the use case and disk

avikivity:
I think 1 month is a sensible default.

gmizrahi:
We use a 20GB instance and had to reduce the retention time to 7 days.
for 8 nodes and 7 days our disk usage is 5.6GB just for the Prometheus data.

It is something which needed to be taken into consideration...

avikivity:
All I can say is, the number of metrics is going to increase all the time.

amnonh:
@tzach please close this issue or merge the pull request

tzach:
fix by #250 
retention is avilable as a command line option 
```
-b storage.local.retention=1000h
```

JackThird:
> fix by #250 retention is avilable as a command line option
> 
> ```
> -b storage.local.retention=1000h
> ```

-b storage.local.retention=1000h didn't work for me.

As per prometheus 2.43 documentation (https://prometheus.io/docs/prometheus/latest/storage/) there are new parameters for retention:
-b "-storage.tsdb.retention.time=7d"
or/and
-b "-storage.tsdb.retention.size=1GB"

PR #177, opened 2017-05-24
Change data retention when using external data to 5 years

pdziepak:
The default data retention for Prometheus is 15 days. That may be all right during normal usage, but doesn't work very well when we want to use it to analyse previously captured data (using the flag `-d`).

When the monitoring containers are started with an external source of data provided the data retention is set to 5 years to avoid this problem.


tzach:
@pdziepak -d is also used when you want to mount an external drive, not just when uploading data from an external source. Maybe we should add a new parameter to set the retention time?

tzach:
related #124

tzach:
Closing this PR, as setting retention is now avilabe via command line option
https://github.com/scylladb/scylla-grafana-monitoring/commit/f1d234a6acc1b9fc520cf6c53db83082789e6170

PR #178, opened 2017-05-29
use a new Scylla logo

tzach:
Use the new Scylla logo from its new URL


tzach:
@eyalgutkind can you please test?

PR #184, opened 2017-06-22
Add option for per shard or per instance charts

tzach:
This PR add 3 new selectors to the per server dashboard:
- Level: Cluster, Instance or shard
- Instance: select which instance to present
- Shard: select which shard present

This allows, for example to select one instance, and break it to its shards

image

### DIFF
--- a/grafana/scylla-dash-per-server.1.7.json
+++ b/grafana/scylla-dash-per-server.1.7.json
@@ -381,7 +381,7 @@
                                 "expr": "avg(scylla_reactor_utilization{instance=~\"[[node]]\", shard=~\"[[shard]]\"} ) by ([[by]])",
                                 "intervalFactor": 1,
                                 "refId": "A",
-                                "legendFormat": "{{instance}} {{shard}}",
+                                "legendFormat": "",
                                 "step": 1
                             }
                         ],
@@ -464,7 +464,7 @@
                                 "intervalFactor": 1,
                                 "metric": "",
                                 "refId": "A",
-                                "legendFormat": "{{instance}} {{shard}} -Requests Served per Server",
+                                "legendFormat": "",
                                 "step": 1
                             }
                         ],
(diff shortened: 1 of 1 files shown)


tzach:
@glommer I think this is in the direction of what you asked for.
I'm not happy with the Legend format in this PR, as they present the word "instance" or "shard" when one select the other options. I'm sure there is a better solution.

glommer:
@tzach maybe I am doing something wrong, but I can't see any data when pointing to a live cluster.

It is working, because the "Total requests" query, which is not being modified, shows data just fine. But the rest of the queries don't really. I am using 1.7 btw - should that be a master only thing ?

tzach:
@glommer retested with Scylla 1.7 AMI
Per node:
image

Per Shard of one node:
image

glommer:
Tzach,

Apparently the 1.7 graph is converted, but 2017.1 isn't, and master also
isn't.

I am also writing a patch here, in the context of issue #190, and I am
having to change all three dashboards.

As I have said before, I strongly think that we should be auto-generating
those graphs instead of having multiple copies of the same thing.

tzach:
> Apparently the 1.7 graph is converted, but 2017.1 isn't, and master also
isn't.

Yes. If you are good with 1.7 we can do teh same for 1.6,  2017.1 and the IO dashboard.


> As I have said before, I strongly think that we should be auto-generating
those graphs instead of having multiple copies of the same thing.

Make sense. Some templating / auto-generating can be useful in theory. We will need to look and the PR to judge.

Few examples I found:
https://github.com/uber/grafana-dash-gen
https://github.com/jakubplichta/grafana-dashboard-builder

glommer:
On Mon, Jul 3, 2017 at 2:46 PM, Tzach Livyatan <notifications@github.com>
wrote:

> Apparently the 1.7 graph is converted, but 2017.1 isn't, and master also
> isn't.
>
> Yes. If you are good with 1.7 we can do teh same for 1.6, 2017.1 and the
> IO dashboard.
>
> As I have said before, I strongly think that we should be auto-generating
> those graphs instead of having multiple copies of the same thing.
>
> Make sense. Some templating / auto-generating can be useful in theory. We
> will need to look and the PR to judge.
>
> Few examples I found:
> https://github.com/uber/grafana-dash-gen
> https://github.com/jakubplichta/grafana-dashboard-builder
>

I'll try to do it myself if I have some time later this week. The second
link you sent looks interesting.

PR #186, opened 2017-06-29
add Grafana admin password option

### LINKED ISSUE #185: Add documentation and flag to use API key for secured Grafana dashboards
The default setting in the load-grafana.sh script is with disabled authentication.
Users with secured Grafana servers will need to enable authentication to the clients.

Please add description and a flag to the script to enable connection to secured Grafana servers

tzach:
This PR add an option to change the Grafana admin password from the command line, by adding a `-a password` to start_all.sh
By adding a password option, one activates the following:
- Enabling basic Grafana auth
- Setting the Admin user password
- Disabling anonymous access to Grafana

Nothing in the UX change when admin password is not provided.

Fix #185

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,6 @@ ___
 ___
 
 
-
-
 #### Connecting Scylla and the Monitoring locally - the local flag
 When running the Prometheus and Grafana on the same host as scylla, use the local `-l` flag, so processes inside the
 containers will share the host network stack and would have access to the `localhost`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -103,6 +103,8 @@ ___
 ___
 
 
+
+
```
glommer: unnecessary churn

README.md
```diff
@@ -115,6 +117,7 @@ containers will share the host network stack and would have access to the `local
 
 ### Use
 Direct your browser to `your-server-ip:3000`
+By default, Grafana authentication is disabled. To enable it and set password for user admin use the `-a` option
```
glommer: set a password

glommer (commented):
patch is ok, two minor aesthetic changes suggested.

PR #191, opened 2017-07-03
correctly display writes currently blocked on dirty

### LINKED ISSUE #190: current blocked on dirty is bogus
We have a graph for current number of requests blocked on dirty but it is using the wrong query.
It shows the total cumulative amount, just without taking the first derivative.

The correct metric is

```
scylla_database_requests_blocked_memory_current
```

glommer:
Fixes #190

### DIFF
--- a/grafana/scylla-dash-per-server.1.7.json
+++ b/grafana/scylla-dash-per-server.1.7.json
@@ -1663,7 +1663,7 @@
                         "steppedLine": false,
                         "targets": [
                             {
-                                "expr": "sum(scylla_database_requests_blocked_memory{instance=~\"[[node]]\", shard=~\"[[shard]]\"}) by (instance)",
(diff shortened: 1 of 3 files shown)


tzach:
Looks good, but I have to admit I do not understand what does "blocked on dirty" means, and what should I do with a high (or low) value.

glommer:
It means that the system is not able to flush its memtables fast enough, so
the system has to throttle. This metric is of the utmost importance to
diagnose latency issues.

In an ideal world, that would be due to disk is saturated. In real life,
sometimes that happens due to the lack of a good CPU isolation (CPU
scheduler). When faced with high values here, one has to first determine if
this is a disk or CPU issue. At that point, the solution depends on the
scenario.

It might be just a non-issue if the workload is expected to be disk bound.
It could mean "get faster disks", or it could mean "lower your parallelism"

tzach:
Thanks for the detailed explanation.
We should somehow propagate the info to the dashboard itself.
There is a Description field we are not using.

PR #196, opened 2017-07-04
Fix non-unique panel IDs

### LINKED ISSUE #194: Some panels don't refresh properly and have distorted layout
For me it's always "Cache Insertions", "Cache evictions" and "Cache Total Bytes". They constantly look as in this screen:

screenshot from 2017-07-04 14-47-28

After I enter the edit mode on them, they start to show data, but also start to have distorted layout:

screenshot from 2017-07-04 14-50-02

Their definition doesn't differ from the panels which refresh ok, other than for labels and query string.

There are errors in the JS console about requests being cancelled:

screenshot from 2017-07-04 14-52-48

Prometheus responds quickly to all requests though.

The panel being "bad" correlates with a panel sharing an id with another panel:
```
$ cat grafana/scylla-dash-per-server.master.json | grep "\"id\":" | sort | uniq -c | sort -r | head
      2                         "id": 48,
      2                         "id": 46,
      2                         "id": 45,
      2                         "id": 44,
```

Assigning unique IDs fixes the problem.

tgrabiec:
Caused refresh and layout problems.

Fixes #194.

### DIFF
--- a/grafana/scylla-dash-per-server.1.7.json
+++ b/grafana/scylla-dash-per-server.1.7.json
@@ -3989,7 +3989,7 @@
                         "content": "<h1 style=\"color:#5881c2; border-bottom: 3px solid #5881c2;\">CQL</h1>",
                         "editable": true,
                         "error": false,
-                        "id": 44,
+                        "id": 144,
                         "isNew": true,
                         "links": [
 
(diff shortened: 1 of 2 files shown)


tzach:
Nice catch @tgrabiec

PR #193, opened 2017-07-04
port per shard to 1.6, master and 2017.1

tzach:
This PR port the per shard option added to 1.7 in #184 to dashboards 1.6, 2017.1 and Master
It also fixes a few minor issues the per share presentation of 1,7

fix #192

PR #197, opened 2017-07-04
Fix query for Cache Total Bytes panel

### LINKED ISSUE #195: The "Cache Total Bytes" panel has incorrect query
It uses `scylla_cache_total{}`, and there is no such metric. It should use `scylla_cache_bytes_total{}`

tgrabiec:
Fixes #195.


tzach:
@tgrabiec both this and #198 looks good, but conflixts with an eariler bigger change #193 
If you can review it first, it would be easier for me to fix the conflict for these two PR.

tzach:
Merged via #199

PR #202, opened 2017-07-13
dots have feelings too.

glommer:
Dots are also valid characters in instance names. The current regex
makes instances with dots in its name fail to show up.


tzach:
Tested. Look good.
Thanks @glommer

PR #206, opened 2017-08-01
Add per shard and per Instance drop-down to all IO dashboards

tzach:


tzach:
@glommer can you please review?

glommer:
@tzach tested and it works well

Thanks!

PR #209, opened 2017-08-10
Add sstables streaming charts

tzach:
Add Active and Queued SSTable streaming chart to the Replica section
image

### DIFF
--- a/grafana/scylla-dash-per-server.2017.1.json
+++ b/grafana/scylla-dash-per-server.2017.1.json
@@ -1,7 +1,6 @@
 {
     "dashboard": {
         "id": null,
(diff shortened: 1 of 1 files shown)


tzach:
@glommer @amnonh is this PR still relevant?

amnonh:
@tzach is this still relevant?

glommer:
It would be still nice to separate the SSTable list into streaming and normal reads, yes.

The graphs would have to be redone from this PR, though.

amnonh:
I've compared the pull request with what we have, it seems that it's the same metrics but in a different order. Do we still need this?

PR #212, opened 2017-08-13
start-all.sh: Forward grafana port option instead of processing it

duarten:
Signed-off-by: Duarte Nunes <duarte@scylladb.com>


tzach:
Thanks @duarten

PR #214, opened 2017-08-14
grafana: add option for HTTP proxy

avikivity:
Often, the Prometheus database will only be open to a few static IPs.
Allow directing grafana to access Prometheus via a proxy (that is presumably
authorized to access Prometheus).

A new option -x proxy:port is provided.

### DIFF
--- a/start-grafana.sh
+++ b/start-grafana.sh
@@ -14,9 +14,9 @@ GRAFANA_ADMIN_PASSWORD="admin"
 GRAFANA_AUTH=false
 GRAFANA_AUTH_ANONYMOUS=true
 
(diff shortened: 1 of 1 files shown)


tzach:
Looks good to me, but I'm not sure how to test it.
@duarten can you please help in the review?

tzach:
@avikivity please add the same parameter to start-all.sh and propogate it to start-grafana.sh
Most users use the first to start the monitoring stack

duarten:
@tzach It looks good to me. I don't think any changes to `start-all.sh` are required since in that case Grafana should have direct access to the Prometheus container.

tzach:
> @tzach It looks good to me. I don't think any changes to start-all.sh are required since in that case Grafana should have direct access to the Prometheus container.

Fair enough, we can add it later if required

PR #228, opened 2017-09-10
Add units to master per server dashboard

tzach:


grafana/scylla-dash-per-server.master.json
```diff
@@ -675,7 +675,7 @@
                         },
                         "yaxes": [
                             {
-                                "format": "short",
+                                "format": "wps",
```
tgrabiec, line 678: scylla_storage_proxy_coordinator_foreground_writes presents current count of such writes, we don't take a time derivative of it, so "wps" (w per second) is not appropriate.

grafana/scylla-dash-per-server.master.json
```diff
@@ -757,7 +757,7 @@
                         },
                         "yaxes": [
                             {
-                                "format": "short",
+                                "format": "rps",
```
tgrabiec, line 760: same comment as for wps.

grafana/scylla-dash-per-server.master.json
```diff
@@ -1008,7 +1008,7 @@
                         },
                         "yaxes": [
                             {
-                                "format": "short",
+                                "format": "wps",
```
tgrabiec, line 1011: it's also a queue length, so not "wps".

grafana/scylla-dash-per-server.master.json
```diff
@@ -1089,7 +1089,7 @@
                         },
                         "yaxes": [
                             {
-                                "format": "short",
+                                "format": "rps",
```
tgrabiec, line 1092: same comment as for wps

grafana/scylla-dash-per-server.master.json
```diff
@@ -4078,7 +4078,7 @@
                         },
                         "yaxes": [
                             {
-                                "format": "short",
+                                "format": "ops",
```
tgrabiec, line 4081: This is right, but we should also fix the long standing bug in the query definition, which should take irate() (time derivative) of the metric

tgrabiec (changes requested):

grafana/scylla-dash-per-server.master.json
tzach: @tgrabiec can you please open an issue for this one?

tzach:
Thanks @tgrabiec latest commit fix the issue you pointed to
Please check

grafana/scylla-dash-per-server.master.json
tgrabiec: There is already an issue for this: https://github.com/scylladb/scylla-grafana-monitoring/issues/158

prometheus/scylla_servers.yml
```diff
@@ -1,4 +1 @@
-# List Scylla end points
-
-- targets:
-  - 172.17.0.2:9180
\ No newline at end of file
+  - targets: ["172.17.0.2:9180"]
```
tgrabiec, line 1: This commit changes scylla_servers.yml, not dashboard definition

tgrabiec (changes requested):

prometheus/scylla_servers.yml
tzach: Thanks, obviously a mistke

amnonh:
This commit is no longer applicable, the file change is no longer in use and there is a different way of specifying the units.
Please open an issue if the problem still exists.

Issue #234, opened 2017-09-23
Remove duplication from dashboards


tzach:
Currently, all 3 dashboards have the same metrics at the top (live nodes, transaction per node, cluster, disk usage). This is both redundant and hard to maintain.

amnonh:
@tzach is this still relevant?

PR #237, opened 2017-09-24
set maximum for load graph higher

### LINKED ISSUE #236: fix units for Load and others
@tzach in commit 33cc238c20aeb51e1f0c3d6be80f80bd952abd70, you changed the load graph to explicitly be maxed out at 100. 

Grafana doesn't like this too much, because when the load is exactly at 100 %, it may coincide with the upper line and is not shown. Hovering the mouse will show that the point is indeed there, and at 100.

Changing this to 101 seems to do the trick: this guarantees that the max line is always above the maximum we can show.

glommer:
We are currently setting the graph for Load at a maximum of 100.  This
change seems to be made for the master graph only, and not for any of
the versions (which reminds me that this week, I haven't said we need to
automate this yet).

The problem with that is that when the load is at exactly 100 %, grafana
may not show the point as it coincides with the upper bar. I was left
out with a graph full of "holes".

Putting the max a bit higher - at 101 - guarantees that the bar is
always shown.

Fixes #236

### DIFF
--- a/grafana/scylla-dash-per-server.master.json
+++ b/grafana/scylla-dash-per-server.master.json
@@ -403,7 +403,7 @@
                             {
                                 "format": "percent",
                                 "logBase": 1,
-                                "max": 100,
+                                "max": 101,
                                 "min": 0,
                                 "show": true
                             },


tzach:
Thank @glommer can you update the other two master dashboards with the same fix?
(at least till we fix #234)

glommer:
@tzach the other dashboards don't have a maximum, so they don't suffer from this. Do you want me to set a maximum for them as well ?

tzach:
> @tzach the other dashboards don't have a maximum, so they don't suffer from this. Do you want me to set a maximum for them as well?

Please do.
Having a constant scale allows you to understand the load at a glance, not forcing you to read the scale. We should use it where we can (not many places)

tzach:
@glommer can we close this PR, now that #240 is merged?

PR #246, opened 2017-10-08
Gentler killing and removing Grafana and Prometheus docker instances 

tzach:
kill and remove Grafana and Prometheus docker instances without error code in case they do not exist


tzach:
@amnonh ping

PR #260, opened 2017-12-21
allow grafana container to be named

glommer:
For people like me, who monitor various clusters, I end up with
a lot of clusters named agraf-3000, agraf-3001, agraf-3002.

Weeks later, I have no idea which is which. I usually resort to
docker rename after the fact, but it doesn't cost us much to add
it as an option here.


tzach:
Thanks @glommer

PR #261, opened 2017-12-28
Make Datasource name configurable.

matt2000:
load-grafana.sh is intended for users with existing Grafana installations, who may already have another datasource named 'prometheus,' making the name unavailable for use by the Scylla prometheus service.

Alternatively, a user may wish to run separate Prometheus services for different scylla clusters, but monitor them in the same Grafana. (This was my particular use case.)

This branch leaves the default behavior of load-grafana.sh unchanged, but adds the option to provide a `-d DATASOURCE_NAME` parameter to allow a different name. The only new dependency is the standard unix `sed` utility which should be available on any system capable of executing the current script.

A more elegant solution than this would be to integrate this feature into make_dashboards.py, but that script is not used by all of the dashboards, and the approach in this branch was easier to implement quickly.

### DIFF
--- a/grafana/scylla-dash-io-per-server.1.5.json
+++ b/grafana/scylla-dash-io-per-server.1.5.json
@@ -48,7 +48,7 @@
                             "rgba(237, 129, 40, 0.89)",
                             "rgba(50, 172, 45, 0.97)"
                         ],
-                        "datasource": "prometheus",
+                        "datasource": "DATASOURCE_NAME",
                         "editable": true,
                         "error": false,
                         "format": "none",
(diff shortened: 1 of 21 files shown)


tzach:
Thanks @matt2000 sorry for the late response 
@amnonh can you please help and review this?

amnonh:
When we started working on the make_dashboard, we thought about adding parameter mapping support (add a parameter mapping/command line) that will replace parameters in the json file. we didn't find a real use case for it then, maybe we can use it for this.

I think a better way would be to add the following functionality to make_dashboard.py
The following need to be done:
1. add a parameter object to make_dashboard.py
 - the parameter will be load from file and can be override from the command line
2. use template resolution in make_dashboard.py that uses the parameter object to replace the parameters.
3. replace in the source in the template file to a parameter.

tzach:
closing this PR, following @amnonh comment above
@amnonh or @matt2000 a new PR using make_dashboard  will be appreciated.

PR #272, opened 2018-01-26
replace `sudo docker` with `docker`

### LINKED ISSUE #270: Remove sudo usage
Each docker invocation is preceded by `sudo` https://github.com/scylladb/scylla-grafana-monitoring/search?utf8=%E2%9C%93&q=sudo&type=

To use docker user should be docker group, `sudo` can be removed.

mmatczuk:
Fixes #270


amnonh:
I have tested the patch and it is working

PR #282, opened 2018-02-08
Add latencies and overview

glommer:
This series add latency metrics to our dashboard - one long requested feature that we can now support, thanks to the storage proxy histograms that we export.

I found that they don't quite fit that well to the per-server dashboard, which is already packed with so much information.

So in preparation to that, I changed the Cluster dashboard into an "overview" dashboard - which is something we needed to have anyway.

### DIFF
--- a/grafana/scylla-dash.2.1.template.json
+++ b/grafana/scylla-dash.2.1.template.json
@@ -165,6 +165,221 @@
             },
             {
                 "class": "row",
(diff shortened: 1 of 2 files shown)


glommer:
A comment: Cheers for the new templating system @amnonh . It was really easy to add those graphs.

tzach:
@amnonh can you please review

PR #291, opened 2018-03-09
grafana: per-server-master: Rework cache dashboard

tgrabiec:
It now shows both per row and per partition metrics.

It also has "reads with misses" and "reads with no misses", which proved to be useful.

Preview:

screenshot from 2018-03-09 16-19-52

### DIFF
--- a/grafana/scylla-dash-per-server.master.template.json
+++ b/grafana/scylla-dash-per-server.master.template.json
@@ -783,12 +783,9 @@
                 "height": "25px",
                 "panels": [
                     {
-                        "class": "text_header_panel",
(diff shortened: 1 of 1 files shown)


tgrabiec:
This change utilizes new metrics added for row-level eviction in commit scylladb/scylla@0ebfe448e3c473352348746966692a652e0364b5

tzach:
Looks good.
@amnonh can you test it, please?

amnonh:
I've tested it and it looks good.
There is still too much redundancy that could be avoided, but it's not limited to this series, so I'll merge it and we should do an additional clean up of the templates files.

PR #302, opened 2018-03-27
Set y-max of the Load chart as 101 percent

### LINKED ISSUE #301: New 2.1 Overview Dashboard has some graphics issue, in the CPU-reactor graph
he missing lines are only on the overview dashboard, while in the per-server Dashboard it looks OK.

image

tzach:
fix #301

### DIFF
--- a/grafana/scylla-dash.2.0.json
+++ b/grafana/scylla-dash.2.0.json
@@ -283,7 +283,7 @@
                             {
                                 "format": "percent",
                                 "logBase": 1,
-                                "max": 100,
+                                "max": 101,
                                 "min": 0,
                                 "show": true
                             },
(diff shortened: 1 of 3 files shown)


tzach:
@amnonh please review

amnonh:
We should probably define a type for graph with percentage and remove the duplication definitions of the y axis, it will be in a single place in types.jon

tzach:
> We should probably define a type for a graph with percentage and remove the duplication definitions of the y axis, it will be in a single place in types.jon

No idea how to do that. Can you please send a new PR for such a fix?

tzach:
@amnonh I'm merging since its a reported bu
Let's do the template definition in a follow-up PR

PR #315, opened 2018-04-18
Using a more generic way to extract Prometheus docker IP

### LINKED ISSUE #311: Prometheus Host missing in Grafana
Just setup monitoring and noticed that it could not connect to Prometheus. 

screen shot 2018-04-10 at 9 29 19 am

Looked at the data sources and noticed that the host is missing.

screen shot 2018-04-10 at 9 27 11 am

After adding the host of the docker container, connectivity was established from Grafana to Prometheus:

screen shot 2018-04-10 at 9 30 37 am

tzach:
fix #311

### DIFF
--- a/start-all.sh
+++ b/start-all.sh
@@ -143,7 +143,7 @@ fi
 
 # Can't use localhost here, because the monitoring may be running remotely.
 # Also note that the port to which we need to connect is 9090, regardless of which port we bind to at localhost.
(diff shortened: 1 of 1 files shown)


tzach:
Got the syntax from @mavenugo (tnx!)  https://github.com/moby/moby/issues/21658#issuecomment-203652928

PR #330, opened 2018-06-21
sort shardlist in numerical order

glommer:
Currently the shard list is not sorted. Users have complained
that if they have 80+ shards it is hard to find the shards they
are interested at.

Sorting helps a lot with that.


tzach:
@amnonh are you OK with this PR?

tzach:
or should we use #331 instead?

amnonh:
I think we should go with #331 it covers both shards and nodes and for more dashboards.

PR #332, opened 2018-06-22
enhance per server dashboard with useful metrics

glommer:
First, to open up some real estate in the dashboard, I am removing the total
operations graph that sits on top: it is now redundant, because the user
can aggregate per Cluster in the entire dashboard.

With this out of the way, I am doing the following:
- Adding a graph with foreground CPU utilization. That is the CPU used by
  request processing, excluding compaction, flushes and other things. The reason
  for that is that users are usually scared of spikes. Even if we tell them that
  spikes are fine because they are the result of isolatable background processes,
  it is hard to *prove* that without further analysis. This graph will help.

- time spent in violations: A lot of the latency issues we have, specially in
  higher percentiles come from task quota violations. We have a metric for this
  now and it will help us correlate latency spikes in time

- Client connections: in the past few months, this is *THE* top metric we
  have been looking at to detect problems. It harms us a lot that it is not
  part of the main dashboard.

In the process of doing the above, I am also doing my best to document the new
graphs. The text will appear in the tooltip in the top left corner of the graph.

Caveat: The request per sec graph, where it is, becomes too narrow. I don't
know how to make it wider.

### DIFF
--- a/grafana/scylla-dash-per-server.master.template.json
+++ b/grafana/scylla-dash-per-server.master.template.json
@@ -51,17 +51,18 @@
                     },
                     {
                         "class": "ops_panel",
-                        "bars": true,
                         "targets": [
                             {
-                                "expr": "sum(irate(scylla_transport_requests_served{}[30s])) + sum(irate(scylla_thrift_served{}[30s]))",
+                                "expr": "sum(irate(scylla_transport_requests_served{instance=~\"[[node]]\", shard=~\"[[shard]]\"}[30s])) by ([[by]]) + sum(irate(scylla_thrift_served{instance=~\"[[node]]\", shard=~\"[[shard]]\"}[30s])) by ([[by]])",
                                 "intervalFactor": 1,
-                                "legendFormat": "Total Requests",
+                                "legendFormat": "",
+                                "metric": "",
                                 "refId": "A",
                                 "step": 1
                             }
                         ],
-                        "title": "Total Requests"
+                        "title": "Requests Served per [[by]]",
+                        "description": "Amount of requests served as the coordinator. Imbalances here represent dispersion at the connection level, not your data model."
                     },
                     {
                         "class": "dashlist",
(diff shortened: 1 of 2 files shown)


glommer:
Example 

ex

grafana/scylla-dash-per-server.master.template.json
```diff
@@ -127,6 +131,39 @@
...
                         ],
                         "title": "Total Storage"
+                    },
+
+                    {
+                        "class": "ns_panel",
+                        "pointradius": 1,
+                        "targets": [
+                            {
+                                "expr": "sum(irate(scylla_scheduler_time_spent_on_task_quota_violations_ns{group=\"main\", instance=~\"[[node]]\", shard=~\"[[shard]]\"}[30s])) by ([[by]])",
```
glommer, line 141: the violation graph shouldn't specify the group. Let me guys know if the rest is okay and I will fix that for next version

glommer: @amnonh let me know what you think

tzach:
@amnonh can you please review

tgrabiec:
Looks good to me

tzach:
* I fail to see data on the new "Foreground CPU Utilization by Instance" and "Time spent in task quota violations by Instance". Do they use metrics added after 2.1?
* In the future, better to add descriptions and a separate PR for easier review
* Very happy with adding new useful metrics, but I'm not sure about the position in the dashboard:
  - users which know and used to the dashboard expected to see the same metrics.  Placing a new, similar chart in an existing place is confusing. 
  - The top of the dashboard should be used by the most relevant and important charts. CPU is utilization definitely one. Foreground CPU Utilization and time spent in violations are, IMHO, advanced metrics which relevant to experts. Maybe better under a "CPU" section at the bottom (below compacton?).

glommer:
On Tue, Jul 3, 2018, 9:02 AM Tzach Livyatan, <notifications@github.com>
wrote:

>
>    - I fail to see data on the new "Foreground CPU Utilization by
>    Instance" and "Time spent in task quota violations by Instance". Do they
>    use metrics added after 2.1?
>
>
Yes they are in 2.2 only. I am not trying to patch the 2.1 dashboard.

>
>    - In the future, better to add descriptions and a separate PR for
>    easier review
>    - Very happy with adding new useful metrics, but I'm not sure about
>    the position in the dashboard:
>       - users which know and used to the dashboard expected to see the
>       same metrics. Placing a new, similar chart in an existing place is
>       confusing.
>       - The top of the dashboard should be used by the most relevant and
>       important charts. CPU is utilization definitely one. Foreground CPU
>       Utilization and time spent in violations are, IMHO, advanced metrics which
>       relevant to experts. Maybe better under a "CPU" section at the bottom
>       (below compacton?).
>
>
I can move the violations, but I think foreground cpu should be in the same
place as load. The graph is added because people keep asking "why do I have
spikes, this is terrible". That graph is intended to show when it is
terrible and when it isn't. The closer it is to load, the better.

tzach:
@glommer I suspect there are ~4 people in the world who knows what Scylla foreground cpu means. Putting it as one of the 3 top metrics does not make sense to me.

What about following the route of the io dashboard and create a CPU dashboard? 
We can drill down to this kind of metrics there. With the upcoming scheduler tags, we have many more CPU metrics coming.

glommer:
There may be 4 people in the world that know that it exists, but once the graph is there, any person in the world can understand what it actually is. The title of the graph could also be changed by "CPU time spent in foreground operations", which it makes it even more clear.

And if it is still not clear, there is a tooltip help now - unlike in all the other graphs.

That said, a CPU dashboard makes sense. I'll leave it up to you guys to do it. Make sure to include this metric.

PR #333, opened 2018-06-28
Move to prometheus 2.3 v1

vladzcloudius:


tzach:
@vladzcloudius what is the effect of upgrading on existing Prometheus data?

slivne:
@amnonh
- can you please check this out - that the upgrade will also retain the data (I understand that vlad did some test - lets double check)
- can you please verify that the language has not changed and that all graphs and queries should work.

amnonh:
I run into https://github.com/coreos/prometheus-operator/issues/830
Which mean there's a problem moving data from 1.8 to 2.x

The functions seem to work fine

slivne:
The issue is closed on prometheus - so whats the solution/workaround ?

vladzcloudius:
On 07/01/2018 01:27 AM, Tzach Livyatan wrote:
>
> @vladzcloudius <https://github.com/vladzcloudius> what is the effect 
> of upgrading on existing Prometheus data?
>

Data formats are not compatible between 1.x and 2.x.
There is a procedure of migrating the existing data from the old server 
to the new one: https://prometheus.io/docs/prometheus/2.0/migration/

tzach:
Thanks @vladzcloudius 
We should document the upgrade process. A good timing to bump this project major release to make this breaking upgrade more visible

tzach:
@vladzcloudius @amnonh can we close this PR?
I understand #344 is replacing this PR

tzach:
> @vladzcloudius @amnonh can we close this PR?
> I understand #344 is replacing this PR
> 

@amnonh ?

PR #385, opened 2018-09-06
README.md: clarify configuration

### LINKED ISSUE #382: Hard-to-understand phrase in README.md
README.md has the following phrase:

**For every server, there are two targets, one under scylla job which is used for the scylla metrics. Use port 9180.**

It was impossible for me to understand what this is supposed to say. What is "server" in this sentence - a Scylla server? prometheus server? A whole cluster? And why "every"? What does "scylla job" refer to, and how do you do something "under" it? I'm guessing that this is some sort of Prometheus terminology? And what is a "target"?

Maybe once I understand what you really meant here I'll come back and suggest different wording. But it would be nice to make this text readable for someone who has absolutely no prior knowledge of graphana or prometheus or anything, and just knows Scylla terminology like cluster or node.

nyh:
I had a hard time understanding the beginning of the configuration section.
It referred to concepts I did not understand, and whichhad not been
previously introduced.

I propose the following rewrite of a couple of paragraphs.

Fixes #382

Signed-off-by: Nadav Har'El <nyh@scylladb.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,12 +73,11 @@ centos $ sudo service docker start
 ```
 
 ### Configuration
-Update `prometheus/scylla_servers.yml` and `prometheus/node_exporter_servers.yml` with the targets (server you wish to monitor).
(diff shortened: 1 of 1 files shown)


tzach:
Looks good to me

PR #414, opened 2018-10-22
Output docker logs when container fails to start

amoskong:


amnonh:
@tzach 
I think a better UX would be a message such as:
run `docker logs aprom` for more information (as an example)

amnonh:
@amoskong let me know if you want to send an updated version, or I will close it

amoskong:
@amnonh I just updated the PR to add a message before dump the docker log. Can you help to review?

amnonh:
@amoskong can you add an example of how it looks like in practice?

amoskong:
```
[amos@amos-centos7 scylla-grafana-monitoring]$ ./start-grafana.sh 
....
Error: Grafana container failed to start
Run `docker logs agraf --details` for more information
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
[amos@amos-centos7 scylla-grafana-monitoring]$ echo $?
1
```

amoskong:
Do you expect that we only note user to check error log by `docker logs xxx --details`, but we don't dump error to output directly?

amnonh:
I think it's better not to dump the logs, specifically as it can get really big and unclear.

Is this something you need for QA?

amoskong:
We use scylla-grafana-monitor project for preparing monitor in testing. I
touched a fail before I created this PR, I saw `Error: Grafana container
failed to start`, but the instance had been removed, I can't check the real
error again.

However, I agreed to just add note without dump the error log.

amnonh:
If you need it for QA, we can add a verbose flag, that would print the logs

tzach:
No need to augment `docker logs` with a verbose mode, its redundant
Adding a message in case of an error make sense, not too noisy.

Another alternative is to adding a link to [monitoring troubleshooting](https://docs.scylladb.com/operating-scylla/monitoring/monitor_troubleshoot/) and give the relevant commands, like `logs` there.

amnonh:
Dropping by default all the logs is a bad idea, the problem is with QA that needs a way to know why a container did not start

amnonh:
Closing, please open an issue with the suggested alternative if you still looking for that functionality

PR #418, opened 2018-10-31
Add a variable for the datasource

glommer:
There are situations in complex deployments where we can be monitoring different
clusters that are in different prometheus servers.

What we have been usually doing in situations like that is having many different
stacks, each with its own grafana/prometheus pair. But it would be useful to be
able to have the same entry point for it.

With this patch, the datasource becomes configurable as a grafana variable. We
can then choose which datasource we'll monitor, here called an "organization"
and seamlessly flip between them.

### DIFF
--- a/grafana/scylla-dash-cpu-per-server.2.3.template.json
+++ b/grafana/scylla-dash-cpu-per-server.2.3.template.json
@@ -146,6 +146,13 @@
                     "name": "shard",
                     "query": "label_values(scylla_reactor_utilization,shard)",
                     "sort": 3
+                },
+                {
+                    "label": "Datasource",
+                    "name": "datasource",
+                    "query": "prometheus",
+                    "refresh": 1,
+                    "type": "datasource"
                 }
             ]
         },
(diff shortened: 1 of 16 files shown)


tzach:
So an organization is a level above clusters, where each org have multiple clusters, and there can be many orgs.
Is it the case?

glommer:
The main difference as I see it is that the organization is in a different
datasource


I find the name organization a bit more pleasant., but if it is confusing I
can name it "Datasource" as well.

tzach:
I think data source make more sense, familiar to Prometheus users

glommer:
ok, I'll change. Any other change you'd like to see done ?

tzach:
Looks fine to me, I will be happy if someone* will smoke test this

`*` @amnonh

glommer:
Branch updated.

tzach:
@amnonh can you please review?

amnonh:
I've tested the patch, I think this should be the first parameter and not the last.
And it's probably worth sorting it

amnonh:
@glommer please see my last comment

amnonh:
@glommer ping

amnonh:
it's no longer applyable, closing

PR #441, opened 2018-12-18
Add materialized view related panels

duarten:
Adds the following MV-related panels:
- Current view update backlog;
- Dropped view updates;
- Currently throttled base writes.

The panels are added to 3.0 and master.

Signed-off-by: Duarte Nunes <duarte@scylladb.com>

### DIFF
--- a/grafana/scylla-dash.3.0.template.json
+++ b/grafana/scylla-dash.3.0.template.json
@@ -367,7 +367,22 @@
                         "title": "",
                         "transparent": true,
                         "type": "text"
+                    },
+                    {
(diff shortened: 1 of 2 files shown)


tzach:
@duarten can you attach screen capture of the dashboard?

duarten:
screen shot 2018-12-18 at 12 32 15

tzach:
Thanks @duarten 
@amnonh is this the right dashboard to include these new panels?

duarten:
I added them to the overview, since they should be looked at together with the "timeouts and errors" panel.

PR #525, opened 2019-03-07
do not specify any particular group when aggregating violations.

### LINKED ISSUE #524: metric for task quota violations is misleading
The current metric says:

```
sum(irate(scylla_scheduler_time_spent_on_task_quota_violations_ns{group=\"main\",instance=~\"[[node]]\" ,cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]\"}[30s])) by ([[by]])
```

This is very wrong, since the main class barely has any problems with task quota violations. Most of the violations come from somewhere else.

We want to see all violations, not just a particular group. So we should not specify `group` here.

glommer:
We are interested in seeing all violations, not just the ones from the main class.
In particular, the main class tends to be very uninteresting. Most of the violations
come from somewhere else.

Fixes #524

### DIFF
--- a/grafana/scylla-dash-cpu-per-server.2.3.template.json
+++ b/grafana/scylla-dash-cpu-per-server.2.3.template.json
@@ -69,7 +69,7 @@
                         "pointradius": 1,
                         "targets": [
                             {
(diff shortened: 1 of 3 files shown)


amnonh:
Don't we need to know the group information? we can add it to the by cluster or add filter parameter

glommer:
No, if we specify no groups it will sum them all, which is what we want for the graphs.

In a deeper analysis we might want to know which group is causing the most stalls so we can look for them in the code, but that's too much information for the default grafana graph.

That graph should tell us "are there stalls?" Right now it almost always answers NO (because they rarely come from main) when in reality we're full of them.

PR #586, opened 2019-04-23
Fix chart legend formats

martin-sucha:
Having the same static string for many graph series is not
very helpful as you can't tell the source instance when you
spot an anomaly in the graph.

Let's format the legend with instance and device for metrics
that are not dependent on `by` template parameter.

For metrics that change the grouping based on the `by` template
parameter we just use empty legend format since the tags that
generate series in the graph vary. With empty legend format,
Grafana includes all the information needed using its default
format.

### DIFF
--- a/grafana/scylla-dash-per-server.2.0.json
+++ b/grafana/scylla-dash-per-server.2.0.json
@@ -2329,7 +2329,7 @@
                             {
                                 "expr": "irate(node_disk_writes_completed{device=\"$monitor_disk\"}[30s])",
                                 "intervalFactor": 1,
-                                "legendFormat": "Disk Writes per Server per Second",
+                                "legendFormat": "{{ instance }} / {{ device }}",
                                 "metric": "",
                                 "refId": "A",
                                 "step": 1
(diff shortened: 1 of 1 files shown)


martin-sucha:
This MR only changes one of the templates, but we probably need to update all of them. Is that correct?

amnonh:
@martin-sucha I don't mind merging it, but this change is for a specific dashboard for scylla 2.0.

Syclla monitoring latest does not update those dashboards.
Later releases uses a template dashboard system and this issue was resolved.

martin-sucha:
I updated the 2.0 dashboard as that's the one I've been using and didn't notice the newer ones don't have such problem. I'll try using the newer ones then. If this is a known bug and was already fixed in the newer dashboards, why keep the old dashboards if they are not updated with bugfixes?

It wasn't obvious to me how the dashboards are generated. Could you please update the README so that it's clear how the dashboards are generated and where outside contributors should look first when adding changes?

amnonh:
You need to use the same version as your scylla version, so if you are using Scylla 2.0, use the 2.0 dashboards.

Just to be clear, this is different than the Scylla Monitoring version (current 2.2 and 2.3 will be released soon).

If you are using Scylla version 2.3 or 3.0, you probably upgrade to the latest Scylla Monitoring, we keep on improving it and adding new dashboards and functionality.

There is a wiki page about updating the dashboard here:
https://github.com/scylladb/scylla-grafana-monitoring/wiki/Updating-a-dashboard-for-fun-and-profit

And we are in a process of adding it to the documentation as part of the monitoring docs:
https://docs.scylladb.com/operating-scylla/monitoring/

amnonh:
BTW why not just clear all the legendFormat? I'm not sure it would look when using aggregation

amnonh:
Closing

PR #593, opened 2019-05-03
add kubernetes operator dashboard generator

jwnx:
The kubernetes operator runs node-exporter on a kubernetes node,
not Scylla node. Therefore, datacenter and cluster information does not
exist nor makes sense. Moreover, data is extracted from a
'kubernetes_node' instead of 'node'.

This patch makes those adaptations given a -k flag for load-grafana.sh
and -ko for make_dashboards.py.

### DIFF
--- a/load-grafana.sh
+++ b/load-grafana.sh
@@ -5,10 +5,11 @@ VERSIONS=$DEFAULT_VERSION
 GRAFANA_HOST="localhost"
 GRAFANA_PORT=3000
 DB_ADDRESS="127.0.0.1:9090"
+KUBERNETES_OPERATOR_BUILD=false
 
-usage="$(basename "$0") [-h] [-v comma separated versions ] [-g grafana port ] [-H grafana hostname] [-m alert_manager ip:port] [-p ip:port address of prometheus ] [-a admin password] [-j additional dashboard to load to Grafana, multiple params are supported] [-M scylla-manager version ] -- loads the prometheus datasource and the Scylla dashboards into an existing grafana installation"
+usage="$(basename "$0") [-h] [-v comma separated versions ] [-g grafana port ] [-H grafana hostname] [-m alert_manager ip:port] [-p ip:port address of prometheus ] [-a admin password] [-j additional dashboard to load to Grafana, multiple params are supported] [-k build dashboards for kubernetes operator] [-M scylla-manager version ] -- loads the prometheus datasource and the Scylla dashboards into an existing grafana installation"
 
-while getopts ':hg:H:p:v:a:j:m:M:' option; do
+while getopts ':hkg:H:p:v:a:j:M:' option; do
   case "$option" in
     h) echo "$usage"
        exit
@@ -22,6 +23,8 @@ while getopts ':hg:H:p:v:a:j:m:M:' option; do
     H) GRAFANA_HOST=$OPTARG
        ;;
     j) GRAFANA_DASHBOARD_ARRAY+=("$OPTARG")
+       ;;
+    k) KUBERNETES_OPERATOR_BUILD=true
        ;;
     p) DB_ADDRESS=$OPTARG
        ;;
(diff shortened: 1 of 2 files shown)


jwnx:
Also would be nice to cherry-pick to `scylla-monitoring-2.3` stable branch

load-grafana.sh
```diff
@@ -49,7 +52,11 @@ for f in scylla-dash scylla-dash-per-server scylla-dash-io-per-server scylla-das
     if [ -e grafana/$f.$v.template.json ]
     then
         if [ ! -f "grafana/build/$f.$v.json" ] || [ "grafana/build/$f.$v.json" -ot "grafana/$f.$v.template.json" ]; then
-            ./make_dashboards.py -t grafana/types.json -d grafana/$f.$v.template.json
+	    if [ "$KUBERNETES_OPERATOR_BUILD" = true ]; then
```
amnonh, line 55: no need for the extra if, just set KUBERNETES_OPERATOR_BUILD to --kubernetes-operator

make_dashboards.py
```diff
@@ -248,11 +249,36 @@ def write_as_file(name_path, result, dir):
...
+            u'class': u'template_variable_all',
+            u'label': u'kubernetes_node',
+            u'name': u'kubernetes_node',
+            u'query': u'label_values(node_disk_writes_completed, instance)'
+            }
+
+    result["dashboard"]["templating"]["list"].append(dict(kubernetes_node_variable))
+
+    for row in result['dashboard']['rows']:
+        if row["class"] == "row":
```
amnonh, line 263: class row is not a must

make_dashboards.py
```diff
@@ -248,11 +249,36 @@ def write_as_file(name_path, result, dir):
...
+           for panel in row['panels']:
+               if 'targets' in panel:
+                   for target in panel['targets']:
+                       if '(node_' in target['expr']:
```
amnonh, line 267: First, this patch is too specific, you are taking metrics specific names (node_) that you know belongs to the node_exporter.
I would rather not do that.
Second, I'm trying to understand, why do you need all that?
wouldn't it all work out of the box if you specify node_exporter file with kubernetes nodes in it?

load-grafana.sh
jwnx: ok

make_dashboards.py
jwnx: For your first point: me too. Open to suggestions.

Second: 
Are you talking about the scrape config? If you are, there is a configuration file for it and it does scrape the `kubernetes_node`variable.

Grafana doesn't use `kubernetes_node` nor cares. I can't override `node` to `kubernetes_node`, because, afaik, you can't have two labels with the same name. The `instance=[[node]]` query won't select anything because `instance` uses kubernetes node's internal IP instead of Scylla's internal IP which won't match the `node` definition (which is Scylla's internal IP). Plus, on the dashboards, `node_` variables seems to care about `dc`, which doesn't exist on kubernetes. 

For example, metrics:
```
node_*(instance=10.0.0.1, kubernetes_node=gke.node1.cluster)
scylla_*(instance=50.0.0.1, dc=as, ....)
```
Label:
```
node=label_target(scylla_*, instance) // so, 50.0.0.x
```
Query:
```
node_*(instance=[[node]], dc=$dc, ...) // nothing
```

What this patch does:
```
node_*(instance=10.0.0.1, kubernetes_node=gke.node1.cluster)
scylla_*(instance=50.0.0.1, dc=as, ....)
```
Label:
```
node=label_target(scylla_*, instance) // so, 50.0.0.x
kubernetes_node=label_target(node_*, kubernetes_node) // so, 10.0.0.x
```
Query:
```
node_*(instance=[[kubernetes_node]],  dc=$dc|$^, ...) // something
```

amnonh:
we split the dashboards so that node_ related metrics will be in their own dashboard this way you can have different ips for the node_exporter and for scylla.

If you are using kubernetes as your node source (instead of reading the ip from a file).
you can use relabel to modify the instance label.

About changing the dc, it's better to change the dashboard to allow empty dc than to do that in a script.

glommer:
> we split the dashboards so that node_ related metrics will be in their own dashboard this way you can have different ips for the node_exporter and for scylla.
> 
> If you are using kubernetes as your node source (instead of reading the ip from a file).
> you can use relabel to modify the instance label.
> 
> About changing the dc, it's better to change the dashboard to allow empty dc than to do that in a script.

I am in agreement with the above.

amnonh:
@jwnx I think we can replace this PR with a change to the dashboards that would allow empty DC.
Ping me if you need help with the configuration and relabeling.
If it will not work we'll see what we can do.

If it does work, it's best to add that to the documentation or knowledge base.

amnonh:
Closing

PR #668, opened 2019-07-03
create external directories if they don't exist

glommer:
I verified today that if the prometheus data directory does not exist,
docker itself will try to create it.  However the directory will then be
created with the wrong permissions and as a result prometheus fails to
start if the code is being executed as a user.

Create the directory ourselves to prevent this issue. I am also doing
the same for grafana since it is bound to have the same issue.


amnonh:
Fixes #669

amnonh:
I don't like the solution on one hand (the general problem still remain, if the application will create a directory it will be created with root permissions), but I couldn't find a better solution on the other hand, so I'm merging.

glommer:
the application will only create it as root if it is run as root -> in which case there is no problem.

Otherwise it will be created as the user who ran start-all.sh, which is what we want.

amnonh:
that's true only for the topmost directory that this patch creates, still, if an application inside the docker container will create a new directory somewhere else, it would face the same issue.

PR #779, opened 2019-11-22
grafana: fix typo in mount path label

ultrabug:
There's a little typo on the label showing on grafana :)


amnonh:
Thanks for the Fix!

PR #820, opened 2020-02-17
Adding path mapping for docker-in-docker run

### LINKED ISSUE #821: Make it possible to run scylla-monitoring from the inside of the docker container
In SCT we start hydra to run scylla-monitoring, in case of docker backend it is going to be executed there there, i.e. from the inside of hydra.
It leads to having bad docker mount points (docker run -v .. ) since docker will use host environment, and container know nothing about them and uses paths from the container environment.

In order to metigate that we need to add docker mapping functionality wich will provide ability for map contianer mounting points path to host path.

dkropachev:
Idea here that when we are running start-all.sh from the inside of the docker it is needed to map all files/folder that are going to be user for -v or --mount for docker run command, in order to get correct/real docker host path

Fixes #821

### DIFF
--- a/.include.sh
+++ b/.include.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+function convert_docker_mapping() {
+  # Supposed to use for mapping docker-in-docker paths for -v or --mount usage 
+  # Usage:
+  #  convert_docker_mapping "/source/path:/dest/path,/source2/path:/dest2/path" "/source/path/file_to_bind"
+  dir_mappings=${1//,/ }
+  target_path=$2
+  result=$target_path
+  for el in $dir_mappings; do
+    src=${el//:*}
+    dst=${el//*:}
+    if [[ "${target_path}" == "${src}" ]]; then
+      result=$dst
+      break
+    elif [[ ${target_path} == "${src}/"* ]]; then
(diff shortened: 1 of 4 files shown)


amnonh:
@dkropachev We need an issue to merge this PR

dkropachev:
> @dkropachev We need an issue to merge this PR

https://github.com/scylladb/scylla-monitoring/issues/821

bentsi:
@dkropachev I am not convinced this functionality should be part of Scylla Monitoring product. Let's discuss

amnonh:
@bentsi @dahankzter Please decide if this PR is still relevant, if not please close it and #821

dkropachev:
> @bentsi @dahankzter Please decide if this PR is still relevant, if not please close it and #821

We have worked this problem around in sct, but problem is still here and probably worth to be addressed.

amnonh:
@bentsi @dahankzter we still have this PR and the issue open.
Can you summarize the discussion if it is still needed?

bentsi:
running monitoring stack from Docker (Docker inside Docker) is not common use case and involves complexity that can be solved by another means (like by giving exactly the same path inside the container as outside). @dkropachev please close this

amnonh:
Closing, it can be re-open if we'll need it

PR #983, opened 2020-06-30
added baseline monitor docs

lauranovich:


tzach:
@lauranovich this wrong approach.
- files under `_build` are generated by Sphinx, no need to add them to git
- We do not need a directory per version. We will generate doc per version from the project  branches / tags

I suggest to start with a minimal project, including just the sphinx part nad one Hello World page.
Once this is working, we should add content.

lauranovich:
ok probably easier to close this and start again

PR #999, opened 2020-07-21
Added gh actions script and replaced _utils folder

dgarcia360:
- Added ``.github/workflows/pages.yml`` from the base theme.
- Replaced ``docs/_utils`` folder to match the base theme files.
- Moved docs under docs/source to match the base theme folder structure.


tzach (approved):

tzach:
LGTM
The doc tree is, for some reason, different than what we have under doc, but we can fix it in a follow up PR

Issue #1000, opened 2020-07-21
Local doc tree is different that original docs


tzach:
The local doc including the following items:

- Scylla Monitor Support Matrix
- Install Scylla Monitor Stack
- The start-all.sh script
- Scylla Monitor Interfaces
- Deploy Scylla Monitor Without Docker
- The CQL Optimiztion
- Cross DC
- Adding and Modifying Dashboards
- Alerting
- Troubleshoot Monitor

This is different than the [docs under Monitoring 3.4](https://docs.scylladb.com/operating-scylla/monitoring/3.4/)

- Install Scylla Monitoring Stack
- The start-all.sh script
- Scylla Monitoring Interfaces
- Deploy Scylla Monitoring Without Docker
- Troubleshoot Monitoring
- Troubleshooting guide for Scylla Manager and Scylla Monitoring integration
- Upgrade Guide - Monitoring
- CQL Optimization Dashboard
- Adding and Modifying Dashboards
- Alerting

Why is it different?
 
![image](https://user-images.githubusercontent.com/170200/88093510-c6447a00-cb9a-11ea-959f-2110c994c118.png)
![image](https://user-images.githubusercontent.com/170200/88093706-128fba00-cb9b-11ea-9a55-650cd6ef8592.png)

tzach:
@lauranovich can we close this?

PR #1040, opened 2020-09-10
Update index.rst

lauranovich:
resorted the index


lauranovich:
FIX https://github.com/scylladb/scylla-monitoring/issues/1000

PR #1045, opened 2020-09-14
Fix version matching

dgarcia360:
This PR adds an additional regex rule to match exact versions for this particular repo version naming convention.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -537,9 +537,9 @@ sitemap_url_scheme = "{link}"
 # -- Options for multiversion --------------------------------------------
 # Whitelist pattern for tags (set to None to ignore all tags)
 # smv_tag_whitelist = r'^.*$'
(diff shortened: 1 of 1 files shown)


amnonh:
I don't mind merging this but I don't understand what this patch is doing and what is it solving

dgarcia360:
@amnonh The docs automated publication stopped working: http://scylladb.github.io/scylla-monitoring/

This is because the current regex ``\b(scylla-monitoring-3.4)\b`` is matching the version ``scylla-monitoring-3.4.2`` instead of the version defined ``scylla-monitoring-3.4``. The regex considers the dot after 4. as if ``scylla-monitoring-3.4`` was the complete word.

image

Other repositories we have automated docs are not facing  this problem because the version naming convention they follow is slightly different (``VERSION-scylla`` instead of ``<repo-name>-<version>``):

image

The new addition ``(?!\S)`` checks if the string matched is followed by a whitespace character, matching the expected version:

image

PR #1108, opened 2020-11-04
Read Latencies: Fix incorrect formula for read average latency

### LINKED ISSUE #1110: Average read latency is misscalculated
The formula for read latency is wrong:
`read_laency_sum/`**`write`**`_latency_count` instead of `read_laency_sum/`**`read`**`_latency_count`

eliransin:
The read average latency formula was calculated
by dividing the read sum divided by *write* count. This
changes the formula to divide by the read count.

Fixes #1110

### DIFF
--- a/grafana/build/ver_2019.1/scylla-overview.2019.1.json
+++ b/grafana/build/ver_2019.1/scylla-overview.2019.1.json
@@ -526,7 +526,7 @@
             "pluginVersion": "7.1.3",
             "targets": [
                 {
-                    "expr": "sum(rate(scylla_storage_proxy_coordinator_read_latency_sum{instance=~\"[[node]]\",cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]|$^\"}[60s]))/(sum(rate(scylla_storage_proxy_coordinator_write_latency_count{instance=~\"[[node]]\",cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]|$^\"}[60s])) + 1)",
+                    "expr": "sum(rate(scylla_storage_proxy_coordinator_read_latency_sum{instance=~\"[[node]]\",cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]|$^\"}[60s]))/(sum(rate(scylla_storage_proxy_coordinator_read_latency_count{instance=~\"[[node]]\",cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]|$^\"}[60s])) + 1)",
                     "instant": true,
                     "intervalFactor": 1,
                     "legendFormat": "",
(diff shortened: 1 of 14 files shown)


amnonh:
@eliransin good catch, can you split the commit between a template change followed by a generated file change?
This allows easier backporting if needed.

eliransin:
> @eliransin good catch, can you split the commit between a template change followed by a generated file change?
> This allows easier backporting if needed.

@amnonh No problem, can you hint me on how to do this? Here is what I understand from your request:
1. First commit only change the files containing .template in the name
2. second commit run some script to generate the other files (which one?)
Thanks

eliransin:
ok - found it :slightly_smiling_face:

eliransin:
@amnonh I have regenerated the dashboard but it seams that they have a lot of diffs now.
I can't tell if this is good or not.
I used: `./generate-dashboards.sh -v 2019.1,2020.1,4.1,4.2,master -F`

amnonh:
I'll have a look, also you should have an issue this patch would close

eliransin:
@amnonh opened an issue and referenced it in the PR prologue.

eliransin:
@amnonh the extra diff comes from dc5553dc7b98e8b364152fd205f99889aba909a6 , it touched templates and didn't generate
the dashboards.

nyh:
> @eliransin good catch, can you split the commit between a template change followed by a generated file change?
> This allows easier backporting if needed.

@amnonh I don't know anything about grafana so maybe I'm asking a stupid question, but why are "generated files" even in the repository? For example in Scylla core, we put the C++ source code, but not the generated object files or executables or whatever generated from them - these are built automatically by "ninja" or "cmake" or whatever. Can't we do the same also in the monitoring project?

amnonh:
@nyh This is actually an excellent question, which has two parts:
1. Why there are generated files in the first place?
We load and manage dashboards using files. Grafana file format is verbose and repetitive, we wanted a way to make it feasible for a human to make changes to a file and track it with git.
I came up with the template mechanism, it works pretty much like HTML and CSS.
I describe it all here: grafana and 
https://www.scylladb.com/2018/03/02/better-grafana-dashboards/

Regular grafana dashboards are still supported, and the dashboard files can be a mixture of templated and not templated
Explanation on how to add and modify dashboard can be found in the documentation:
https://docs.scylladb.com/operating-scylla/monitoring/3.4/updating_dashboard/

2. Why adding the generated results to the repository:
We started without it and generated it for the users, but it adds complexity to using the monitoring.
Following other similar libraries, we add the generated files. (think about all the front end libraries or plugins with a `dist` directory)

The question is do we need to add it to master?
We could have kept master without the generated files (I'm thinking about moving to that in the future). But Master is in use by development and QA quite frequently, so it's still there.

nyh:
>     1. Why adding the generated results to the repository:
>        We started without it and generated it for the users, but it adds complexity to using the monitoring.

Why? To start the monitoring you use a script like start_all.sh. This script can generate the files if they don't yet exist, or even better - regenerate them if changed (as in "make") or even every time (it probably takes a fraction of second, no?)

> The question is do we need to add it to master?
> We could have kept master without the generated files (I'm thinking about moving to that in the future). But Master is in use by development and QA quite frequently, so it's still there.

But the result can be what happened now - that someone can forget to update the generated files...

PR #1128, opened 2020-11-10
alternator: Adding alternator streams to dashboard

fruch:
Now that alternator streams is in master, we want to see those
metrics in alternator dashboard

### DIFF
--- a/grafana/alternator.master.template.json
+++ b/grafana/alternator.master.template.json
@@ -650,6 +650,176 @@
                 "panels": [
                     {
                         "class": "collapsible_row_panel",
+                        "title": "Streams"
+                    }
+                ]
+            },
+            {
+                "class": "header_row",
+                "panels": [
+                    {
+                        "class": "plain_text",
+                        "content": "<h1 style=\"color:#5881c2; border-bottom: 3px solid #5881c2;\">Streams Actions</h1>"
+                    }
+                ]
+            },
+            {
+                "class": "row",
+                "panels": [
+                    {
+                        "class": "ops_panel",
+                        "span": 4,
+                        "targets": [
+                            {
+                                "expr": "$func(rate(scylla_alternator_operation{instance=~\"[[node]]\",cluster=~\"$cluster|$^\", dc=~\"$dc\", shard=~\"[[shard]]|$^\", op=\"ListStreams\"}[60s])) by ([[by]])",
+                                "intervalFactor": 1,
+                                "legendFormat": "",
+                                "refId": "A",
+                                "step": 1
+                            }
+                        ],
+                        "title": "ListStreams by [[by]]"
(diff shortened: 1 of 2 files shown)


amnonh:
Hi @fruch please commit the changes to the template as a separate commit and the generated file as a separate commit.
and if you can add a screenshot of how it looks like after your change

fruch:
> Hi @fruch please commit the changes to the template as a separate commit and the generated file as a separate commit.
> and if you can add a screenshot of how it looks like after your change

image

fruch:
@amnonh now it's in two commit as you asked

fruch:
FYI, @elcallio @nyh, with this we could see it in alternator dashboard

amnonh:
@fruch one more thing, it should close an issue with a **Fixes** 
Is that relevant to 4.2 or only to master?

fruch:
We noticed this today in a meeting, it wasn't raised in any issue.

It's only for master, 
I guess it would be a nice to have for 4.3 (even that you don't have a version for it yet)

nyh:
@fruch is right - Scylla 4.2 (which was branched long ago...) did not yet have Alternator Streams. So 4.3 will be the first version where these new metrics will become available.

amnonh:
@nyh @fruch because the dashboard is becoming crowded, are those metrics useful to the users?

grafana/alternator.master.template.json
```diff
@@ -942,6 +1112,29 @@
                         }
                     ],
                     "query": "GetItem,PutItem,UpdateItem,DeleteItem"
+                },
```
amnonh, line 1115: I'm not sure it does what you want it to do. This is a template variable with a single value, if you only have one static value (GetRecords), no need to make it a variable.

If there are multiple values you want to support, they are missing.

If this is a preparation for future extension, maybe it's premature, but ok

fruch:
> @nyh @fruch because the dashboard is becoming crowded, are those metrics useful to the users?

This is hard to tell now, since we don't have users yet for this feature, it's not yet GA.
it's helpful to us, and since it's an public API, same as GetItem/PutItem and such. it's reasonable to be able to show it.

it might be o.k. to have it collapsed by default, or put in a separate dashboard.

thing is we don't yet have feedback from users on this feature.

grafana/alternator.master.template.json
fruch: I was trying to follow the pattern that was there before, to minimize my work. 
not that I know of more APIs we'll need have latency metric for, again this is still early to tell which would be mostly needed and if we'll add more metrics to the APIs.

amnonh:
I'll merge it as is, we don't have many Alternator users, We will need to give more thought to the Alternator dashboard in the future

PR #1163, opened 2020-12-09
fix typo

lauranovich:
needs to be backported to other versions(3.5.1 for example)
fix https://github.com/scylladb/scylla-doc-issues/issues/660

### DIFF
--- a/docs/source/cql_optimization.rst
+++ b/docs/source/cql_optimization.rst
@@ -1,6 +1,6 @@
-===================
-The CQL Optimiztion
-===================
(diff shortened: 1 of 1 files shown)


lauranovich:
@amnon - this needs to be backported to versions 3.5.1 and up

lauranovich:
@amnonh - commmit https://github.com/scylladb/scylla-monitoring/commit/31ec10e6d119562596663bd1808d817a4ff66466 needs to be applied as well

PR #1176, opened 2020-12-16
change latest version to correct version number

lauranovich:
update latest version so that the warning message is updated


lauranovich:
I am not sure if this needs to be backported, it may need to be

amnonh:
BTW @lauranovich if you add a Fixes note, it would automatically closes the issue for you

PR #1327, opened 2021-03-17
prometheus.rules: fix summary typos and query expressions syntax

ultrabug:
with this commit we:

- fix some English typos
- normalize CQL term in upper case
- normalize query expression spaces / syntax


ultrabug:
@amnonh hope you like it <3

PR #1359, opened 2021-04-06
add version table for Grafana and Prometheus.

### LINKED ISSUE #1241: need a compitability matrix for grafana prometheus to monitoring version
For users that use their own servers, we need to specify what versions they need to use

lauranovich:
FIXES #1241


amnonh:
Wouldn't it be better to be order in descending order? typically users are looking for the latest version(s)

tzach (approved):

PR #1469, opened 2021-06-29
new metrics table for LWT

lauranovich:
FIXES https://github.com/scylladb/scylla-docs/issues/2584


amnonh:
@tzach it's your call, but I don't think metrics should be documented in the monitoring system, metric are scylla specific and should be tied to a scylla version (or enterprise).

amnonh:
@tzach ping, please see my previous comment

lauranovich:
Moving to Scylla - not sure how considering we don't have much on the metrics
Closing

PR #1474, opened 2021-07-07
change version to correct version number

lauranovich:


lauranovich:
no need for this one, it can be done from master

PR #1500, opened 2021-08-03
fix high latency alerts expression

sbhattiprolu:
match the record rule for read write latencies for high latency alerts


amnonh:
Good catch, can you open an issue that this pr will close? (with a `Fixes` comment)
also from the alert comment, it seems that this should be the p95 recording rules and not the average

sbhattiprolu:
Fixes #1504

@amnonh  the description for the alert is hidden in the UI, please open it, you may be seeing the `description` for the prior alert in that list of alerts. 


image

PR #1563, opened 2021-10-05
adds sidebar to new design so users can choose the version

lauranovich:
Adds the sidebar back to the HP. This allows users to choose the version


lauranovich:
@dgarcia360 is this what you had in mind?

PR #1601, opened 2021-11-25
Update monitor_without_docker.rst

GavinJE:
Grammar correction.


amnonh:
I'm merging as it's a grammer issue, but maybe it could use some re-wording.

PR #1724, opened 2022-05-10
add the "_common" folders to the exclude patterns

annastuchlik:
I've added the __common_ folders to the exclude patterns so that they are skipped by search.


annastuchlik:
@tzach Can you review this PR?

tzach:
LGTM

PR #1738, opened 2022-05-17
fix using GF_USERS_DEFAULT_THEME to set the default theme

fruch:
`GF_USERS_DEFAULT_THEME=.*` was insted quotes, hence wasn't
treated by bash as regex


fruch:
@amnonh found when trying to switch to 4.0 in https://github.com/scylladb/scylla-cluster-tests/pull/4750

PR #1791, opened 2022-09-16
Update CURRENT_VERSION.sh to 4.0.1

mrVanboy:
This is a fixup for the 4.0.1 release that reports itself as 4.0.0.


mrVanboy:
Not sure if it's a good idea to re-release `4.0.1` and maybe we should rather set version to the `4.0.2` and release it as `4.0.2`

amnonh:
just changing the version is not enough, you need to recreate the dashboards.
I will take care of it

mrVanboy:
Thank you. Feel free to close this PR 🙂

PR #1799, opened 2022-10-02
start-all.sh: add --archive switch to disable Prometheus TTL

avikivity:
When loading an old archive, there is a risk that it is older than the retention time and so Prometheus will delete it instantly. This can be worked around with a `-b "-storage.tsdb.retention.time=30y"` switch, but that is too much to remember and to type.

Provide a --archive switch to help out people who mostly work with archives.


amnonh:
I've created #1800 so it will be included in the release notes

PR #1808, opened 2022-10-21
Update docker-compose.example.yml

KadoBOT:


tzach:
@KadoBOT can you please explain why and what this change is?
Both PR and commit message inclued no info.

PR #1869, opened 2023-01-31
Fix fallback rules for SG-specific latency panels in scylla-details.

michoecho:
The panels are supposed to be SG-specific, but their fallback rules forgot to take SGs into account. (The primary rules, based on Prometheus recoding rules, are fine).
Fix that.

### DIFF
--- a/grafana/scylla-detailed.template.json
+++ b/grafana/scylla-detailed.template.json
@@ -1693,7 +1693,7 @@
                         "span": 3,
                         "targets": [
                             {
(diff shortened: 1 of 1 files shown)


amnonh:
@michoecho First we should have a monitoring issue to go with the PR.
Second, I think we should drop the second part and relay on recording rules only.

And if there is a problem with the recording rules, fix that

amnonh:
Replaced by #1873

PR #1870, opened 2023-01-31
Fix the `all_scheduling_group` recording rule

michoecho:
- The rule is supposed to filter out inactive scheduling groups, but it only looks at writes, not reads. In particular, it filters out active read-only groups.
- The rule is missing from the back-fill file.
- The rule does not take `cluster` into account, so it displays scheduling groups from other clusters.

Fix all of the above.

### DIFF
--- a/prometheus/prom_rules/back_fill/3.8/rules.1.yml
+++ b/prometheus/prom_rules/back_fill/3.8/rules.1.yml
@@ -153,6 +153,8 @@ groups:
     expr: histogram_quantile(0.95, sum(rate(scylla_storage_proxy_coordinator_cas_write_latency_bucket{}[60s])) by (cluster, le, scheduling_group_name))
     labels:
       by: "cluster"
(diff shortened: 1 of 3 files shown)


amnonh:
Fixes #1872

prometheus/prom_rules/back_fill/3.8/rules.1.yml
```diff
@@ -153,6 +153,8 @@ groups:
     expr: histogram_quantile(0.95, sum(rate(scylla_storage_proxy_coordinator_cas_write_latency_bucket{}[60s])) by (cluster, le, scheduling_group_name))
     labels:
       by: "cluster"
+  - record: all_scheduling_group
```
amnonh, line 156: this is a new recording rules, for the back-filling, it should be under 4.2 (ie. we need a 4.2 directory with that rule in it)

Issue #1498, opened 2021-08-03
Remove reverse order read warnings once Scylla work on it is done


amnonh:
There is a work in Scylla to improve the performance of reverse order read.
Once this is done, the warning shuld be remved from the Monitoring

bhalevy:
Refs #1413

PR #2004, opened 2023-06-09
doc: update info about reversed queries

annastuchlik:
Refs: https://github.com/scylladb/scylla-doc-issues/issues/831

This commit removes the page that describes reverse queries, as well as fixes the info about reverse queries on the CQL Optimization page.

Similar changes are introduced in the ScyllaDB core docs with this PR: https://github.com/scylladb/scylladb/pull/14190


amnonh:
We currently have a warning when a user uses reverse order queries, that warning link to the page you remove.
@tzach do we want to remove it all together?
If not, we need to update the page, not remove it

tzach:
Since ScyllaDB 5.0 Reverse queries are less than an issue. IMHO, removing the warning from both monitoring and docs is safe.

amnonh:
Relates to #1498

PR #2061, opened 2023-09-15
doc: remove dead links from upgrade guides

annastuchlik:
This PR removes dead links to ScyllaDB Monitoring Stack from upgrade guides.

The links were added when the Monitoring Stack upgrade guides were part of the core ScyllaDB documentation. Now they are both dead and unnecessary, so they were removed.

At a minimum, this PR must be backported to branch-4.5, but it should also be backported to branch-4.4 (to fix the current docs for that version).


annastuchlik:
Thanks! Please backport to branch-4.4 and branch-4.5 (otherwise it won't be visible to users).

amnonh:
Already backported

PR #2100, opened 2023-10-26
prometheus.rules.yml: make InstanceDown alerts 'error'

vladzcloudius:
InstanceDown is definitely not a warning - it's an error that needs to be treated immediately.

### DIFF
--- a/prometheus/prom_rules/prometheus.rules.yml
+++ b/prometheus/prom_rules/prometheus.rules.yml
@@ -133,7 +133,7 @@ groups:
     expr: up{job="scylla"} == 0
     for: 30s
     labels:
-      severity: "warn"
+      severity: "error"
     annotations:
       description: '{{ $labels.instance }} has been down for more than 30 seconds.'
       summary: Instance {{ $labels.instance }} down
@@ -141,7 +141,7 @@ groups:
     expr: sum(up{job="scylla"}>0)by(instance) unless sum(scylla_transport_requests_served{shard="0"}) by(instance)
     for: 1m
     labels:
-      severity: "warn"
+      severity: "error"
     annotations:
       description: '{{ $labels.instance }} instance is shutting down.'
       summary: Instance {{ $labels.instance }} down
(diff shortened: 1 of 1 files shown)


amnonh:
Fixes #2101

@vladzcloudius there are a few different scenarios here, If a node is in shuting down mode, it means it's on purpose, why do you think it should be an error and not a warning?

vladzcloudius:
> Fixes #2101
> 
> @vladzcloudius there are a few different scenarios here, If a node is in shutting down mode, it means it's on purpose, why do you think it should be an error and not a warning?

Nodes are not supposed to be down in general.
If anything is "on purpose" then alerts should be explicitly silenced in advance.
And if they are not silenced - the Infra Operator needs to get informed about it and act ASAP. Hence "error".

amnonh:
> > Fixes #2101
> > @vladzcloudius there are a few different scenarios here, If a node is in shutting down mode, it means it's on purpose, why do you think it should be an error and not a warning?
> 
> Nodes are not supposed to be down in general. If anything is "on purpose" then alerts should be explicitly silenced in advance. And if they are not silenced - the Infra Operator needs to get informed about it and act ASAP. Hence "error".

When you replace a node you have a node that is technically in the cluster but is down, I don't think this should be an error, we do send a warning

vladzcloudius:
> > > Fixes #2101
> > > @vladzcloudius there are a few different scenarios here, If a node is in shutting down mode, it means it's on purpose, why do you think it should be an error and not a warning?
> > 
> > 
> > Nodes are not supposed to be down in general. If anything is "on purpose" then alerts should be explicitly silenced in advance. And if they are not silenced - the Infra Operator needs to get informed about it and act ASAP. Hence "error".
> 
> When you replace a node you have a node that is technically in the cluster but is down, I don't think this should be an error, we do send a warning

Again, when you replace a node you need to silence corresponding alerts.
@amnonh, this is a very important change. 
Please, consider merging.

PR #2114, opened 2023-11-06
prometheus.rules.yml: restrict tooManyFiles rule

### LINKED ISSUE #2113: AlertManager: TooManyFiles generates false alarms for a partition with the OS
Scylla-Monitoring Version: 4.5.0

*Description*
tooManyFiles alert is checked for every partition on the node. 
However we in fact only care about the one that has scylla's `data` folder.
By default it's `/var/lib/scylla` (and the directory in question is `/var/lib/scylla/data`).

We probably want the default Alert Manager configuration check only that partition.
Users that have a non-default mount options configuration would be able to adjust this rule.

vladzcloudius:
Checking this alert for all present partitions is nice however we saw that the one that has OS files often hits some of thresholds and start creating a noise.

If fact we only care about the partition that has scylla 'data' directory and by default this partition is /var/lib/scylla.

Let's restrict these alerts (there are 3 with different thresholds and corresponding severities) to the mount point above only.

Any user that has a different mount point hosting scylla's 'data' directory will have to adjust the filtering correspondingly.

Fixes #2113

### DIFF
--- a/prometheus/prom_rules/prometheus.rules.yml
+++ b/prometheus/prom_rules/prometheus.rules.yml
@@ -262,26 +262,26 @@ groups:
       description: 'OOM Kill on {{ $labels.instance }}'
       summary: A process was terminated on Instance {{ $labels.instance }}
   - alert: tooManyFiles
-    expr: (node_filesystem_files - node_filesystem_files_free) / on(instance) group_left count(scylla_reactor_cpu_busy_ms) by (instance)>20000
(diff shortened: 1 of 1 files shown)


vladzcloudius:
cc @ebenzecri @igorribeiroduarte

amnonh:
@vladzcloudius I don't mind merging it, as it does not impact the cloud, but keep in mind that we did encounter nodes that ran out of root disk space, which is why we are looking at all partitions.

PR #2139, opened 2023-12-25
grafana: fix by variable

### LINKED ISSUE #2138: 4.6.0 - By Instance, doesn't work on any of the dashboards
*Installation details*
Panel Name: All panel that can show data by instance
Dashboard Name: Os Metric (but it apply to all other dashboards)

Scylla-Monitoring Version: 4.6
Scylla-Version: 2024.1-rc2

all of the dashboards that need to show data per instance, so all of the data together: 
image

example of such run:
```
hydra investigate show-monitor 065fe747-1a6d-4f96-8dad-2b0648201bd3
``

here is example of such query:
```
sum(node_memory_MemAvailable_bytes{instance=~"[[node]]",cluster=~"$cluster|$^", dc=~"$dc"}) by ([[by]])/sum(node_memory_MemTotal_bytes{instance=~"[[node]]",cluster=~"$cluster|$^", dc=~"$dc"}) by ([[by]])
```
![image](https://github.com/scylladb/scylla-monitoring/assets/340979/d85ad58e-4ef8-499d-b287-da676dde4a34)


replacing it `[[by]]` with `instance` 
```
sum(node_memory_MemAvailable_bytes{instance=~"[[node]]",cluster=~"$cluster|$^", dc=~"$dc"}) by ([[by]])/sum(node_memory_MemTotal_bytes{instance=~"[[node]]",cluster=~"$cluster|$^", dc=~"$dc"}) by (instance)
```

![image](https://github.com/scylladb/scylla-monitoring/assets/340979/d5d3d0a7-d765-48ab-b81b-6451cf162bcd)

fruch:
seems like custom variable is not full controlled by the query parameter, and you need to feed it a comma seperated with `key : value` for it to work as expected, and the rest of the data we feed in `option` does really work anymore

Closes: #2138

### DIFF
--- a/grafana/scylla-advanced.template.json
+++ b/grafana/scylla-advanced.template.json
@@ -829,7 +829,7 @@
                             "value": "instance,shard"
                         }
                     ],
-                    "query": "Cluster,DC,Instance,Shard",
+                    "query": "Instance : instance,Cluster : cluster,DC : dc,Shard : instance\\,shard",
                     "type": "custom"
                 },
                 {
--- a/grafana/scylla-os.template.json
+++ b/grafana/scylla-os.template.json
@@ -613,7 +613,7 @@
                             "value": "instance,cpu"
                         }
                     ],
-                    "query": "Cluster,DC,Instance",
+                    "query": "Instance : instance,Cluster : cluster,DC : dc,Shard : instance\\,cpu",
                     "type": "custom"
                 },
                 {


amnonh:
@fruch I'm not against this PR, I just don't understand why you get a different behavior than the one I'm getting, what browser and OS are you using?

amnonh:
@fruch and please add a ```Fixes``` with an issue number, this pr closes

fruch:
> @fruch I'm not against this PR, I just don't understand why you get a different behavior than the one I'm getting, what browser and OS are you using?

it's ubuntu and I'm using both Chrome and Firefox, and it's the same in both.

amnonh:
Replaced by #2140 there was a file missing, you didn't manage to add it to this PR so I just add a new PR

PR #2149, opened 2024-01-10
set DOCKER_HOST only if needed

fruch:
we have places where we use a specific DOCKER_HOST in order to be able to support case of running the monitoring stack inside a running podman instance via socket, those places changing the DOCKER_HOST are breaking this usage, and were preventing the monitoring stack from starting

### DIFF
--- a/start-datadog.sh
+++ b/start-datadog.sh
@@ -62,17 +62,19 @@ for val in "${ENV_ARRAY[@]}"; do
         ENV_COMMAND="$ENV_COMMAND -e $val"
 done
 
-if [ ! -z "$is_podman" ]; then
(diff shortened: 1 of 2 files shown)


amnonh:
a PR should fix an issue, can you open one?

fruch:
It's in the context of this issue

https://github.com/scylladb/scylla-cluster-tests/issues/7080

amnonh:
@fruch I've backported to branch 4.6, this means for QA it is already available

fruch:
> @fruch I've backported to branch 4.6, this means for QA it is already available

thanks mean while, I'm patching it at SCT end, for people to be able to retrieve monitor data that didn't yet had this fix (we are caching the monitoring stack with the data)

PR #2190, opened 2024-02-28
DOCS: fix make_dashboards.py prerequisite

Rikkuru:
make_dashboards.py uses python3 and requires pyyaml


amnonh:
it's only for development, users do not need python

PR #2197, opened 2024-03-05
Add cloud-image build scripts

benipeled:
This PR adds cloud-image scripts and a Packer template to build scylla-monitor AMI and GCE images,

@amnonh please go through the files under the `packer/files/` folder and let me know if we have leftovers that can be removed.

Ref https://github.com/scylladb/scylla-pkg/issues/3268


packer/files/add_centos_user.sh
```diff
@@ -0,0 +1,31 @@
...
+            shift 1
+            ;;
+        *)
+            print_usage
+            ;;
+    esac
+done
+
+if [ "$OS" = "ubuntu" ]; then
+    sudo useradd -m -s $(which bash) -G sudo -G docker centos
```
mykaul, line 25: such a confusing name for a user...

packer/files/conf-from-user.py
```diff
@@ -0,0 +1,153 @@
...
+from pathlib import Path
+
+HOME_DIR='/home/centos/'
+USER='centos'
+dry_run=None
+verbose=None
+
+"""
+{
+"version": "master,5.1",
```
mykaul, line 20: unsure what 5.1 is doing here, but it's not supported anymore.

packer/files/conf-from-user.py
```diff
@@ -0,0 +1,153 @@
...
+    try:
+        req = urllib.request.Request(url, headers=headers)
+        response = urllib.request.urlopen(req)
+        return parse_message(response)
+    except Exception as err:
+        print(f"Unexpected {err=}, {type(err)=}")
+        return None
+
+def getAWSData(args):
+    url = 'http://169.254.169.254/latest/user-data' if args.address == "" else args.address
```
mykaul, line 82: this will fail with IMDSv2 - unsure when it'll be the default and enforced as such in AWS.

packer/files/add_centos_user.sh
benipeled: The whole add-centos-user logic is a terrible cloud-related-workaround that should be removed, but now the focus is on copying the existing process, later we will sort out all this spaghetti

packer/files/install_python
```diff
@@ -0,0 +1,37 @@
...
+if [ "$OS" = "ubuntu" ]; then
+    sudo apt-get install -y python3-setuptools
+    sudo apt-get install -y python3-pip
+    exit 0
+fi
+
+yum -y groupinstall 'Development Tools'
+yum -y install openssl-devel bzip2-devel libffi-devel
+yum remove -y python3
+curl -O https://www.python.org/ftp/python/3.8.10/Python-3.8.10.tgz
```
mykaul, line 27: https://www.python.org/downloads/release/python-3818/ perhaps?

mykaul: (I'm somewhat surprised whatever comes with the OS is not good enough, btw)

packer/files/node-exporter.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Node Exporter
+
+[Service]
+ExecStart=/usr/bin/node_exporter --collector.interrupts
```
mykaul, line 5: why do we need to collect interrupts here?

packer/files/install_python
amnonh: i don't remember which of the ubuntu installation (AWS or gcp) didn't have python installed.

packer/files/node-exporter.service
amnonh: it was taken from Scylla, we can drop it, there will be as work, as @benipeled pointed out to do some cleanup, the code here suffered from long commit cycles and didn't get the attention it need

amnonh:
@benipeled There will be lots of cleaning up to do, but it will be easier to do with follow-up patches.
Much of the work is unified the cloud-solution and the general monitoring.
Once merged, I'll work on it with quicker and smaller cycles of removing files.

benipeled:
> @benipeled There will be lots of cleaning up to do, but it will be easier to do with follow-up patches. Much of the work is unified the cloud-solution and the general monitoring. Once merged, I'll work on it with quicker and smaller cycles of removing files.

Great, so if no other comments please merge
From my side, the next step is adding workflow to auto-build following a new release

packer/files/rule_config.yml
```diff
@@ -0,0 +1,96 @@
...
+  # Apply inhibition if the alertname is the same.
+  equal: ['alertname']
+
+
+receivers:
+- name: 'team-X-mails-regular'
+  email_configs:
+  - from: SENDER_EMAIL
+    to: TARGET_EMAIL_DESTINATION
+    smarthost: SENDER_SMTP_HOST
```
fruch, line 77: @benipeled

who is supposed to fill those correctly ? 

and why are the image override this file, with this broken version of the configuration  ?

i.e. I'm trying to use the latest image, and I altermanager fails cause of this file

amnonh: Siren does, there's an open issue for them to fix that. until they do, I keep a copy of the original file, so it will be possible to override it back and start the stack (this is what I do when I test)

PR #2208, opened 2024-03-07
Add github-actions workflow for building cloud-images

benipeled:
The workflow installs the required packer plugins, authenticate with AWS and GCP and use the packer template to build scylla-monitor cloud-images,

Once done, a file with the cloud-images info will be published to the workflow run, and to the release,

The workflow will auto-trigger following a new release (excluding pre-releases), and can be manually triggered against a specific tag.

Ref: https://github.com/scylladb/scylla-pkg/issues/3268

### DIFF
--- a/.github/workflows/build-cloud-images.yaml
+++ b/.github/workflows/build-cloud-images.yaml
@@ -1,0 +1,75 @@
+name: Build Cloud Images
+
+on:
+  release:
+    types: [released]
+  workflow_dispatch:
+    inputs:
+      monitor_version:
+        description: 'Git tag to build from'
+        required: true
+
+jobs:
+  build-cloud-images:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: 'write'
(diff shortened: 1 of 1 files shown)


benipeled:
Secrets should be added to make the auth stages works

mykaul:
I wonder if we should combine both GCP and AWS builds to the same GH action, or separate into 2 different workflows. Unsure (what about ARM64, btw?)

benipeled:
> I wonder if we should combine both GCP and AWS builds to the same GH action, or separate into 2 different workflows. Unsure (what about ARM64, btw?)

We might want to do it in case we have issues with specific provides but for now, the flow is pretty straightforward, once we have customers/cloud requests or use cases that will require split we'll adjust accordingly

benipeled:
The new workflow verified with a manual run against master - https://github.com/scylladb/scylla-monitoring/actions/runs/8221997175

PR #2262, opened 2024-04-17
prometheus.rules.yml: increase a BF alert threshold to 0.2

### LINKED ISSUE #2263: The bloom filter alert cause too many false positive
The current threshold of the bloom filter alert causes too many false positive alerts. It should be set to a higher value

vladzcloudius:
0.1 was too low and created too many false alarms.
Fixes #2263 2263


vladzcloudius:
cc @mkeeneyj

amnonh:
Merging, thanks for the contribution. In the future, please try and have a PR close an issue if possible (I've added one in this case)

PR #2279, opened 2024-05-07
[cloud image] Copy AMI to testing regions

yaronkaikov:
Same has Scylla images, we need monitoring image to be available also in testing regions so QA can use them

### DIFF
--- a/packer/scylla-monitor-template.json
+++ b/packer/scylla-monitor-template.json
@@ -6,6 +6,7 @@
       "ami_name": "{{ user `monitor_image_name` }}",
       "source_ami": "{{user `aws_source_ami`}}",
       "instance_type": "{{user `aws_instance_type`}}",
+      "aws_regions": "us-west-2,eu-west-2,eu-west-1,eu-central-1,eu-north-1",
       "user_data_file": "files/user_data.txt",
       "subnet_filter": {
         "filters": {


roydahan (approved):

roydahan:
@fruch FYI

packer/scylla-monitor-template.json
```diff
@@ -6,6 +6,7 @@
       "ami_name": "{{ user `monitor_image_name` }}",
       "source_ami": "{{user `aws_source_ami`}}",
       "instance_type": "{{user `aws_instance_type`}}",
+      "aws_regions": "us-west-2,eu-west-2,eu-west-1,eu-central-1,eu-north-1",
```
benipeled, line 9: @fruch @roydahan is it required on all these regions? can't we use one specific?

yaronkaikov: IIUC should be identical to what QA is using for testing regions

fruch: We are using multiple regions, some cases are multi region

And we fall back to other regions, or we won't have enough resources, like spots.

Some instance types are available in specific regions.

So yeah, we need them all

fruch (approved):
LGTM 


Do we have some common labels that can help identify those ? And the version it's using ? I don't want it based on the names, it's doesn't work that good for scylla images

yaronkaikov:
> LGTM
> 
> Do we have some common labels that can help identify those ? And the version it's using ? I don't want it based on the names, it's doesn't work that good for scylla images

We didn't configure any tags for those images, we probably should add some
please define which tags you wish to use, i can add it in a separate commit

fruch:
> > LGTM
> > Do we have some common labels that can help identify those ? And the version it's using ? I don't want it based on the names, it's doesn't work that good for scylla images
> 
> We didn't configure any tags for those images, we probably should add some please define which tags you wish to use, i can add it in a separate commit

1) on tag to identify it as monitoring - can't think of a good name nor value - maybe `type: scylla-monitoring` ?
2) version or branch, or both
3) git sha 

I think those should suffice for most of our needs

benipeled (approved):

yaronkaikov:
> > > LGTM
> > > Do we have some common labels that can help identify those ? And the version it's using ? I don't want it based on the names, it's doesn't work that good for scylla images
> > 
> > 
> > We didn't configure any tags for those images, we probably should add some please define which tags you wish to use, i can add it in a separate commit
> 
> 1. on tag to identify it as monitoring - can't think of a good name nor value - maybe `type: scylla-monitoring` ?
> 2. version or branch, or both
> 3. git sha
> 
> I think those should suffice for most of our needs

This is what we have today 
image


And for master
image

amnonh:
@yaronkaikov When I tried to create an image locally, I get the following errors:
```
Error: Failed to prepare build: "amazon-ebs"

1 error occurred:
        * unknown configuration key: '"aws_regions"'


```

yaronkaikov:
> @yaronkaikov When I tried to create an image locally, I get the following errors:
> 
> ```
> Error: Failed to prepare build: "amazon-ebs"
> 
> 1 error occurred:
>         * unknown configuration key: '"aws_regions"'
> ```

Sorry, it should be `ami_regions` fixed in https://github.com/scylladb/scylla-monitoring/pull/2283

PR #2303, opened 2024-05-27
docs: update the installation instructions for 4.7.*

benipeled:
@amnonh once merged, it should be backported to the 4.7 branch (unless the docs are master base only)

### DIFF
--- a/docs/source/upgrade/upgrade-guide-from-monitoring-3.x-to-monitoring-4.y.rst
+++ b/docs/source/upgrade/upgrade-guide-from-monitoring-3.x-to-monitoring-4.y.rst
@@ -23,6 +23,16 @@ Install 4.y (The new version)
 
 Replace “y” with the new minor and patch release number, for example, 4.0.0.zip
 
+
+.. note::
+    For versions 4.7.* and above, use the format `4.7.*.zip`, for example:
+
+    .. code-block:: bash
+
+        wget -L https://github.com/scylladb/scylla-monitoring/archive/4.7.1.zip
+        unzip 4.7.b.zip
+        cd scylla-monitoring-4.7.b/
+
 Setting the server's files
 --------------------------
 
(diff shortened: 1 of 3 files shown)


amnonh:
The weird thing, I already fixed it

docs/source/upgrade/upgrade-guide-from-monitoring-3.x-to-monitoring-4.y.rst
```diff
@@ -23,6 +23,16 @@ Install 4.y (The new version)
 
 Replace “y” with the new minor and patch release number, for example, 4.0.0.zip
 
+
+.. note::
+    For versions 4.7.* and above, use the format `4.7.*.zip`, for example:
+
+    .. code-block:: bash
+
+        wget -L https://github.com/scylladb/scylla-monitoring/archive/4.7.1.zip
```
amnonh, line 32: you can change the wget command to use the -O tag, so it will work regardless of the version number

docs/source/upgrade/upgrade-guide-from-monitoring-4.x-to-monitoring-4.y.rst
```diff
@@ -24,6 +24,15 @@ Install 4.y.b (The new version)
 
 Replace “y” with the new minor release number, for example, 4.0.1.zip
 
+.. note::
+    For versions 4.7.* and above, use the format `4.7.*.zip`, for example:
+
+    .. code-block:: bash
+
+        wget -L https://github.com/scylladb/scylla-monitoring/archive/4.7.1.zip
```
amnonh, line 32: you can change the wget command to use the -O tag, so it will work regardless of the version number

docs/source/upgrade/upgrade-guide-from-monitoring-3.x-to-monitoring-4.y.rst
benipeled: I'm not sure I got the point here - with `-O` we can specify a file-name-target but the URL-suffix does matter for this case, `scylla-monitoring-x.y.z` vs `x.y.z`,

amnonh: it's not a huge change, just make the second line simpler, I wander if we can use the same parameters for the latest version.

annastuchlik:
Looks good. It must be backported, otherwise the instructions will show an incorrect version number.

annastuchlik (approved):

PR #2406, opened 2024-10-07
scylla-overview: Add SCT test support

k0machi:
This commit adds SCT metrics onto the Overview dashboard, migrating them
from the old dump stored in the SCT repo and instead putting them into
the template with optional "sct_tests" dashproduct flag, allowing them
to be conditionally added to the dashboard.

Task: https://github.com/scylladb/qa-tasks/issues/1444
Closes #2419

### DIFF
--- a/grafana/scylla-overview.template.json
+++ b/grafana/scylla-overview.template.json
@@ -8,21 +8,21 @@
                 "class": "row",
                 "panels": [
                     {
-                      "collapsed": false,
-                      "datasource": null,
-                      "id": "auto",
-                      "gridPos": {
-                        "h": 1,
-                        "w": 24
-                      },
-                      "panels": [],
-                      "title": "Cluster overview $cluster",
-                      "type": "row"
+                        "collapsed": false,
+                        "datasource": null,
+                        "id": "auto",
+                        "gridPos": {
+                            "h": 1,
+                            "w": 24
+                        },
+                        "panels": [],
+                        "title": "Cluster overview $cluster",
+                        "type": "row"
                     }
                 ]
             },
             {
-                "class" : "small_stat_rows"
+                "class": "small_stat_rows"
             },
             {
                 "class": "row",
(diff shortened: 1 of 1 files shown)


amnonh:
Putting the entire JSON dump inside a template file defeats the purpose of a template file.
Do you want it as a separate dashboard? or do you want it to override the overview dashboard?

k0machi:
Still somewhat of a WIP but metrics have been ported:
image

grafana/scylla-overview.template.json
```diff
@@ -8,125 +8,1193 @@
...
+                            {
+                                "expr": "{__name__=~'nemesis(.*)(?:gauge)(.*)'}",
+                                "intervalFactor": 2,
+                                "refId": "B"
+                            }
+                        ],
+                        "title": "Requests Served per [[by]]",
+                        "type": "timeseries"
+                    },
+                    {
```
amnonh, line 226: I prefer to move this part to types.json i.e. create a new class for annotation list

grafana/scylla-overview.template.json
```diff
@@ -8,125 +8,1193 @@
...
+                        "span": 6,
+                        "title": "Other(YCSB/Scylla-bench) Stress tools latency 95%",
+                        "targets": [
+                            {
+                                "expr": "avg(sct_ycsb_read_gauge{type=\"p90\"}) by (instance)",
+                                "format": "time_series",
+                                "interval": "15s",
+                                "intervalFactor": 1,
+                                "legendFormat": "YCSB READ [{{instance}}]",
+                                "refId": "G"
```
amnonh, line 781: why the frefid not starting from A?

amnonh:
The PR includes formatting over the entire document. It first combines tabs and spaces, and even without them, you should do it in a different patch.

Besides the specific type that will be best put in types.json, as it's a lot of rows that are  only relevant to QA, you can define the the few rows inside types.json so the amount of unrelated code inside the overview file will be limited (look at small_stats_row for a reference)

PR #2437, opened 2024-12-26
node_exporter: upgrade to latest version, remove unneeded collectors

mykaul:
This PR update node_exporter to v1.8.2 (just like Scylla) and removes the unneeded collectors that are enabled by default (just like Scylla). The only exception is interrupts - which are still enabled by default by Scylla and are not needed here.

### DIFF
--- a/packer/files/node-exporter.service
+++ b/packer/files/node-exporter.service
@@ -2,7 +2,7 @@
 Description=Node Exporter
 
 [Service]
-ExecStart=/usr/bin/node_exporter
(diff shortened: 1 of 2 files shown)


yaronkaikov:
@mykaul if we have the monitor here, can we drop it in the Scylla container?

mykaul:
> @mykaul if we have the monitor here, can we drop it in the Scylla container?

It's unrelated. This one is the node monitor for the monitoring node. (can it be more confusing?).

PR #2722, opened 2025-11-27
Add script to pre-pull Traefik image for AMI build

defbin:
This pull request introduces a small utility script (`pull_traefik_image.sh`) that automates pulling the Traefik Docker image from multiple registries with retries, fallback logic, and clear error messages.

The script is intended for AMI-building workflows. By pre-caching the Traefik image inside the AMI, we reduce external pull requests during instance startup and avoid rate-limit or availability issues across container registries.

Impact
No changes to existing runtime behavior. Enhances reliability of AMI builds and decreases dependency on external registry pull quotas.


packer/files/pull_traefik_image.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Check if docker command is available
+if ! command -v docker &> /dev/null; then
```
mykaul, line 5: nit: some may prefer Podman - especially when we build on Fedora?

mykaul:
General Q: how is it different than all other containers?

defbin:
> General Q: how is it different than all other containers?

I do not understand the question. Can you provide more context?

amnonh:
@mykaul, this is a cloud-specific thing; it will be part of the image but not part of the start-all.sh script

packer/files/pull_traefik_image.sh
amnonh: this will only run as part of the image creation, we know it's docker, and doker will always be available

mykaul:
> > General Q: how is it different than all other containers?
> 
> I do not understand the question. Can you provide more context?

We have the Loki, Grafana, Prometheus containers and maybe others that we preload during the AMI creation, no? Why does traefik need its own script?
CC @amnonh

defbin:
> Why does traefik need its own script?

This PR addresses a specific issue - https://github.com/scylladb/siren/issues/12346#issuecomment-3584762757. Other containers are out of scope.
There is also a script that checks for the Traefik image in the local cache before pulling it.

To support this for other containers, we would need to add the pre-pull here and add local-image checks in other places. No one has requested that.

mykaul:
I'm surprised we don't have that yet for the rest.

packer/files/pull_traefik_image.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "ERROR: docker command not found. Please install Docker first." >&2
+    exit 1
+fi
+
+TAG="${TAG:-v3.5.6}"
```
carlosms, line 10: Do we have any plan to upgrade the version?
Is this something we should do manually before each monitor release? Or maybe we should fetch here the latest stable version?

defbin: I'm against using dynamic tags like `latest` or `stable` because they are black boxes with unknown issues. The script still needs to be invoked somewhere, and the caller can already specify any version to pull by the script. Here is just one specific version.

adambabik:
@amnonh Thanks for reviewing and merging quickly. When can we expect the new Monitor AMI to test this?

PR #2788, opened 2026-02-05
packer: fix GCP Ubuntu image family name

yaronkaikov:
The image family 'ubuntu-minimal-2404-lts' does not exist in GCP. Changed to 'ubuntu-minimal-2404-lts-amd64' which is the correct image family name for Ubuntu 24.04 LTS minimal in ubuntu-os-cloud.

Fixes: https://scylladb.atlassian.net/browse/RELENG-250

### DIFF
--- a/packer/scylla-monitor-template.json
+++ b/packer/scylla-monitor-template.json
@@ -143,7 +143,7 @@
 
     "gcp_project_id": "scylla-images",
     "gcp_zone": "europe-west1-b",
-    "gcp_source_image_family": "ubuntu-minimal-2404-lts",
+    "gcp_source_image_family": "ubuntu-minimal-2404-lts-amd64",
     "gcp_image_storage_location": "europe-west1",
     "gcp_instance_type": "n1-standard-1",
     "gcp_ssh_username": "ubuntu",


packer/scylla-monitor-template.json
```diff
@@ -143,7 +143,7 @@
 
     "gcp_project_id": "scylla-images",
     "gcp_zone": "europe-west1-b",
-    "gcp_source_image_family": "ubuntu-minimal-2404-lts",
+    "gcp_source_image_family": "ubuntu-minimal-2404-lts-amd64",
     "gcp_image_storage_location": "europe-west1",
     "gcp_instance_type": "n1-standard-1",
```
mykaul, line 148: Unrelated, but that should have been n2!

packer/scylla-monitor-template.json
```diff
@@ -143,7 +143,7 @@
 
     "gcp_project_id": "scylla-images",
     "gcp_zone": "europe-west1-b",
-    "gcp_source_image_family": "ubuntu-minimal-2404-lts",
+    "gcp_source_image_family": "ubuntu-minimal-2404-lts-amd64",
```
mykaul, line 146: We don't have an AMI for ARM, btw?

packer/scylla-monitor-template.json
```diff
@@ -143,7 +143,7 @@
 
     "gcp_project_id": "scylla-images",
     "gcp_zone": "europe-west1-b",
```
mykaul, line 145: Unrelated, but why Europe?

packer/scylla-monitor-template.json
mykaul, line 90: unrelated, but (1) is stop enough, or we need to mask / disable it?
(2) we had in Scylla base image more services we've disabled, such as the silly firmware upgrade service, etc.

WE NEED TO USE THE SAME BASE IMAGE.

packer/scylla-monitor-template.json
amnonh: It's not that important; it's only used during image creation

packer/scylla-monitor-template.json
amnonh: No, never had the need for it. We can test one

PR #2839, opened 2026-04-12
scylla-cql: Use recording rules for repeated query/reads denominators

mykaul:
## Summary
- Add 2 recording rules that pre-compute the total query rate and total reads rate per (cluster, dc, instance, shard)
- Replace 6 redundant `sum(rate(scylla_query_processor_queries...))` and 4 redundant `sum(rate(scylla_cql_reads...))` denominator evaluations in the CQL dashboard

### Recording rules (`prometheus/prom_rules/prometheus.latency.rules.yml`)
Added:
```yaml
- record: cql:query_processor_queries_rate
  expr: sum(rate(scylla_query_processor_queries[60s])) by (cluster, dc, instance, shard)
- record: cql:reads_rate
  expr: sum(rate(scylla_cql_reads[60s])) by (cluster, dc, instance, shard)
```

### Dashboard (`grafana/scylla-cql.template.json`)
Updated 10 expressions that used these metrics as denominators:
- Non-Token Aware percentage (line 812)
- CQL ANY CL percentage (line 1006)
- CQL ALL CL percentage (line 1038)
- Multi partition BATCH percentage (line 1072)
- CQL ONE CL percentage (line 1125)
- CQL QUORUM CL percentage (line 1157)
- Non-Paged CQL Reads percentage (line 780)
- Reversed CQL Reads percentage (line 857)
- ALLOW FILTERING percentage (line 889)
- Range Scans percentage (line 974)

## Notes
- The recording rules use `[60s]` window which matches `$__rate_interval` at the default scrape interval (15-20s). At non-default scrape intervals, there may be minor numerical differences.
- Per-(cluster, dc, instance, shard) dimensions are preserved so dashboard node/shard filtering continues to work.

## Testing
1. **`promtool check rules --lint=none`** passes (122 rules found)
2. **`generate-dashboards.sh -v 2024.2`** succeeds without errors
3. **All generated JSON files are valid**
4. **6 `cql:query_processor_queries_rate` and 4 `cql:reads_rate` references** found in generated output

### Manual testing checklist
- [ ] Deploy updated rules and dashboards to a test environment
- [ ] Verify Non-Token Aware percentage panel renders correctly
- [ ] Verify all CL percentage panels (ANY, ALL, ONE, QUORUM) render correctly
- [ ] Verify Multi partition BATCH percentage panel renders correctly
- [ ] Verify Non-Paged CQL Reads, Reversed CQL Reads, ALLOW FILTERING, and Range Scans panels render correctly
- [ ] Filter by specific node — verify percentages still compute correctly (per-node denominators)
- [ ] Filter by specific shard — verify percentages still compute correctly (per-shard denominators)
- [ ] Compare values with an unmodified dashboard on the same cluster

### DIFF
--- a/grafana/scylla-cql.template.json
+++ b/grafana/scylla-cql.template.json
@@ -777,7 +777,7 @@
                         "description": "All requests should be paged\n\nNon Paged request sources:\n- Client modifying the fetch size\n\nNon Paged requests require reading all the results and returning them in a single request.",
                         "targets": [
                             {
-                                "expr": "100 * ((sum(rate(scylla_cql_unpaged_select_queries{instance=~\"[[node]]\",cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}[$__rate_interval]))-sum(rate(scylla_cql_unpaged_select_queries_per_ks{ks=\"system\",instance=~\"[[node]]\",cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}[$__rate_interval])))/sum(rate(scylla_cql_reads{instance=~\"[[node]]\",cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}[$__rate_interval]))) OR vector(0)",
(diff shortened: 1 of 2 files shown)


mykaul:
Note to reviewer: this one needs to be verified: the recording rules use [60s] window which matches $__rate_interval at the default scrape interval (15-20s). At non-default scrape intervals, there may be minor numerical differences.

amnonh:
@mykaul I'm not against using recording rules, I do that a lot, but there needs to be a reason, and a PR should close an issue.
I'm currently leaning towards no

mykaul:
> @mykaul I'm not against using recording rules, I do that a lot, but there needs to be a reason, and a PR should close an issue.
> I'm currently leaning towards no

I need to open a parent issue. I've asked AI to go over all rules and suggest optimizations. It came up with some, I ruled out some and left 4 of them which seemed reasonable.

amnonh:
@mykaul  Recording rules come with a price, and you need a reason to do it, if you open such an issue, it should describe a problem

PR #2848, opened 2026-04-23
package_metrics_for_support.sh

vreniers:
Introduces the initial version of the script which collects Prometheus metrics for analysis by support. 

The script pulls a lightweight container, which executes against a specific Prometheus endpoint. 
The container comes pre-bundled with a set of queries that are executed. 
The output is a `.smi` output file, that is essentially a `.tar.gz, which bundles a JSON file with the query results.

Tested against several monitoring installations, and environments (MacOS, Ubuntu), as well as rootful and rootless installations.

### DIFF
--- a/package_metrics_for_support.sh
+++ b/package_metrics_for_support.sh
@@ -1,0 +1,282 @@
+#!/usr/bin/env bash
+#
+# Collects performance metrics from your Prometheus instance and packages them
+# into a single metrics export file you can share with ScyllaDB Support.
+#
+# Licensed under the Apache License, Version 2.0. The full license text and
+# the licenses of the bundled dependencies are available inside the container
+# image at:
+#   /app/LICENSE      — this project's license
+#   /app/NOTICE       — attribution notices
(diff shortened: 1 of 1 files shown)


package_metrics_for_support.sh
```diff
@@ -0,0 +1,282 @@
...
+set -euo pipefail
+
+# Versioning. Each release of this script is paired with a specific container
+# image tag: by default the script pulls that exact tag, so a given copy of
+# the script always produces results from a known, reproducible build.
+# Override at run-time with --image-tag <tag> (e.g. --image-tag latest, or a
+# pinned version like --image-tag v1.1) when support asks you to.
+SCRIPT_VERSION="v1.0"
+DEFAULT_IMAGE_TAG="$SCRIPT_VERSION"
+SCRAPER_IMAGE_REPO="public.ecr.aws/scylladb-sre/scylladb-metrics-packager"
```
mykaul: This is not great. Who's maintaining this container? Why is it under a different org than ScyllaDB? @ruthea - let's fix this, please.

PR #2868, opened 2026-07-02
packer: make monitor images public and add branch tag

Annamikhlin:
2 changes:
- Make scylla-monitoring AWS images (and their snapshots) public during release.
- Add a `branch` tag (AWS) and label (GCE) with the version line (e.g. branch-4.16), derived from the release version in the workflow.

The `scylladb-monitor-version` tag (AWS: 4.16.0) and label (GCE: 4-16-0) remain unchanged.

Today each release requires a manual edit in SCT, which pins the monitor image by its exact name. With a stable `branch` tag, SCT can instead select the latest image of a release line.

Fixes: https://scylladb.atlassian.net/browse/RELENG-179


yaronkaikov (approved):
LGTM

PR #2887, opened 2026-08-20
start-all.sh: allow setting the Scylla targets from the command line

avikivity:
Until now the Scylla targets could only come from a file, either prometheus/scylla_servers.yml, ./scylla_servers.yml, a file given with -s, or a target directory. For a quick, ad-hoc monitoring session writing a file just to list a couple of addresses is a nuisance.

Add a --servers option that takes a comma separated list of addresses, each with an optional port, and generates the target file under prometheus/build/ (per stack), so it takes the place of prometheus/scylla_servers.yml. It is rejected together with --target-directory and -L, which provide the targets by other means.


mykaul:
Someday this 1000+ lines shell script would move to Python...

amnonh:
> Someday this 1000+ lines shell script would move to Python...

The start part of the stack has no Python dependency; I hope to keep it that way.

PR #2897, opened 2026-08-31
advanced: add "Host CPU" panels

tgrabiec:
They display `node_cpu_seconds_total` metrics annotated with the "shard" label. So standard aggregation and filtering still works. This allows us to inspect CPU metrics for a given shard.

### How it works ###

`node_cpu_seconds_total` is per (instance, cpu) and carries no shard label, so host CPU time could not be filtered or grouped the way the rest of the dashboard is, and lining a stalling shard up with the steal or softirq time on its own core meant reading two graphs side by side while knowing the shard-to-cpu mapping by heart. The mapping is not guessable: on a 16 core box scylla --smp=4 puts shards 0-3 on CPUs 0, 2, 4 and 6, one per physical core.

`scylla_location_shard` supplies that mapping. Joining on (instance, cpu) with group_left(shard) copies the shard label onto the host counter, so the result filters and aggregates like any other metric here - the [[shard]] filter is applied to the location metric, which is the only side that has a shard label to filter on.

### What is added ###

One row, one panel per mode: user, system, iowait, irq, softirq, steal, nice and idle. The panels aggregate by the dashboard's own [[by]] variable, so cluster, dc and instance all roll up as they do everywhere else, and the panel titles and descriptions stay neutral about granularity because the grouping already states it.

### Grouping ###

The Shard option of [[by]] now groups by instance, shard and cpu rather than instance and shard. Without cpu in the grouping the aggregation drops it and the reactor's CPU never reaches the legend, which is half the point of these panels. Every other metric on this dashboard is scylla_* and has no cpu label, so for them cpu is empty and the grouping is unchanged - verified by comparing scylla_reactor_utilization grouped both ways.

### Interval ###
 
The panels set a 30s min step on the query. scylla-monitoring scrapes node_exporter every 1m but scylla every 20s, and $__rate_interval is max($__interval + min step, 4 * min step); with no min step Grafana substitutes the data source scrape interval and the window resolves to 1m, where rate() of a 1m-scraped "node" counter sees a single sample and returns nothing. A 30s min step makes it 2m. The setting has to sit on the query, not in the panel's query options - the latter only feeds $__interval.

Screenshot From 2026-09-01 00-34-00

### DIFF
--- a/grafana/scylla-advanced.template.json
+++ b/grafana/scylla-advanced.template.json
@@ -454,6 +454,94 @@
                 ]
             },
             {
+                "title": "Host CPU",
+                "collapse": true,
+                "rows": [
+                    {
+                        "class": "row",
+                        "panels": [
+                            {
+                                "class": "percentunit_panel",
+                                "span": 3,
+                                "spec/title": "$func host CPU user by [[by]]",
+                                "spec/description": "Host CPU time in user space. On a pinned reactor this is dominated by scylla itself.\n\nnode_cpu_seconds_total joined to scylla_reactor_location on (instance, cpu). Only CPUs that run a scylla reactor are included.",
+                                "spec/data/spec/queries/0/spec/query/spec/interval": "30s",
+                                "spec/data/spec/queries/0/spec/query/spec/expr": "$topbottom([[filter_limit]], $func(rate(node_cpu_seconds_total{mode=\"user\", instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\"}[$__rate_interval]) * on (instance, cpu) group_left(shard) scylla_reactor_location{instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}) by ([[by]]))",
+                                "spec/data/spec/queries/0/spec/query/spec/legendFormat": "{{cluster}} {{dc}} {{instance}} shard {{shard}} cpu {{cpu}}"
+                            },
+                            {
+                                "class": "percentunit_panel",
+                                "span": 3,
+                                "spec/title": "$func host CPU system by [[by]]",
+                                "spec/description": "Host CPU time in the kernel: syscalls the reactor made, plus kernel work charged to that CPU.\n\nnode_cpu_seconds_total joined to scylla_reactor_location on (instance, cpu). Only CPUs that run a scylla reactor are included.",
+                                "spec/data/spec/queries/0/spec/query/spec/interval": "30s",
+                                "spec/data/spec/queries/0/spec/query/spec/expr": "$topbottom([[filter_limit]], $func(rate(node_cpu_seconds_total{mode=\"system\", instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\"}[$__rate_interval]) * on (instance, cpu) group_left(shard) scylla_reactor_location{instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}) by ([[by]]))",
+                                "spec/data/spec/queries/0/spec/query/spec/legendFormat": "{{cluster}} {{dc}} {{instance}} shard {{shard}} cpu {{cpu}}"
+                            },
+                            {
+                                "class": "percentunit_panel",
+                                "span": 3,
+                                "spec/title": "$func host CPU iowait by [[by]]",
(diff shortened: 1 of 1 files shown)


tgrabiec:
@amnonh I wonder if there is a better way to deal with this 30s interval. Can we use some constant for the node exporter scrape period?

grafana/scylla-advanced.template.json
```diff
@@ -453,6 +453,94 @@
                     }
                 ]
             },
+            {
```
amnonh, line 456: With the new format, you can add a header row
. This row and these panels are based on an unsupported metric, so you should add ```dashversion``` label with the version this will be supported from.
For example, if it's part of 2026.3, add ```"dashversion": [">2026.3"],```

grafana/scylla-advanced.template.json
```diff
@@ -453,6 +453,94 @@
...
+                                "spec/data/spec/queries/0/spec/query/spec/legendFormat": "{{cluster}} {{dc}} {{instance}} shard {{shard}} cpu {{cpu}}"
+                            },
+                            {
+                                "class": "percentunit_panel",
+                                "span": 3,
+                                "spec/title": "$func host CPU system by [[by]]",
+                                "spec/description": "Host CPU time in the kernel: syscalls the reactor made, plus kernel work charged to that CPU.\n\nnode_cpu_seconds_total joined to scylla_location_shard on (instance, cpu). Only CPUs that run a scylla reactor are included.",
+                                "spec/data/spec/queries/0/spec/query/spec/interval": "30s",
+                                "spec/data/spec/queries/0/spec/query/spec/expr": "$topbottom([[filter_limit]], $func(rate(node_cpu_seconds_total{mode=\"system\", instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\"}[$__rate_interval]) * on (instance, cpu) group_left(shard) scylla_location_shard{instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}) by ([[by]]))",
```
amnonh: * it's ok to pick a large enough interval (like 2m)
* for a fast-changing function like CPU, maybe (I'm not sure) you want to use irate (usually I'm against it, but in this case maybe), 
* Note that this row is only applicable when looking at shard view; if you use instance view, with a sum aggregation function, you don't get what you probably want.
* Once we're safely working with Grafana 13, we can do better. Once this is merged, please open an issue about this row to handle the aggregation function better

grafana/scylla-advanced.template.json
```diff
@@ -453,6 +453,94 @@
...
+                            },
+                            {
+                                "class": "percentunit_panel",
+                                "span": 3,
+                                "spec/title": "$func host CPU iowait by [[by]]",
+                                "spec/description": "Host CPU time idle with outstanding I/O.\n\nnode_cpu_seconds_total joined to scylla_location_shard on (instance, cpu). Only CPUs that run a scylla reactor are included.",
+                                "spec/data/spec/queries/0/spec/query/spec/interval": "30s",
+                                "spec/data/spec/queries/0/spec/query/spec/expr": "$topbottom([[filter_limit]], $func(rate(node_cpu_seconds_total{mode=\"iowait\", instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\"}[$__rate_interval]) * on (instance, cpu) group_left(shard) scylla_location_shard{instance=~\"[[node]]\", cluster=\"$cluster\", dc=~\"$dc\", shard=~\"[[shard]]\"}) by ([[by]]))",
+                                "spec/data/spec/queries/0/spec/query/spec/legendFormat": "{{cluster}} {{dc}} {{instance}} shard {{shard}} cpu {{cpu}}"
```
amnonh, line 488: It's not the end of the world, but in instance, in DC or in cluster view, you will see empty labels with no numbers. It's not something that needs to be changed unless you decide that you only want the new row to appear in shard view (I'm not sure this is the case; as with the right aggregation function, it makes sense to look at the max or avg of CPU per instance, for example).

tgrabiec:
https://github.com/scylladb/seastar/pull/3662 was merged to seastar, so undrafting this.

amnonh:
@tgrabiec I've seen that you pushed a new branch, but I don't see that my comments were addressed

tgrabiec:
> @tgrabiec I've seen that you pushed a new branch, but I don't see that my comments were addressed

I don't see any comments.

amnonh (changes requested):
I've added a few comments, the only one that must be included is the dashversion.
maybe we want to add to the row steal time, I saw that we report it

amnonh:
@tgrabiec, it was stacked midway. I see that at least one comment is marked as outdated

amnonh:
@tgrabiec I can't merge without at least the ```dashversion``` fix

PR #2904, opened 2026-09-22
node_exporter_install: symlink /usr/bin/node_exporter

defbin:
Fixes MONITOR-95.

4.16.0 moved node_exporter to `/opt/node_exporter/node_exporter` and left nothing at `/usr/bin/node_exporter`, breaking consumers that reference the old absolute path. Symlink it back.

The symlink is safe: the `scylla-node-exporter` deb owns nothing under `/usr/bin`, and it is still extracted rather than registered in dpkg, so the conflict that motivated the move stays fixed.

The existing "already installed" guard is unchanged: a non-forced re-run still refuses (NODE_EXPORTER_BIN exists), and a `/usr/bin/node_exporter` that points elsewhere still blocks a non-forced install. The new link is run once (`--version`) so a broken link fails the image build.


zgrim:
If we do it here, at the source, then we can drop it directly in /usr/bin and not need the symlink.
IAE, the symlink is fine too.

amnonh:
@defbin did you test it?

packer/files/node_exporter_install
```diff
@@ -69,7 +71,9 @@ if __name__ == '__main__':
     if not is_nonroot() and os.getuid() > 0:
         print('Requires root permission.')
         sys.exit(1)
-    if NODE_EXPORTER_BIN.exists() or (bindir_p() / 'node_exporter').exists() or (bindir_p() / 'prometheus-node_exporter').exists():
+    # Symlinks excluded: the compat symlink below is not a prior install.
+    legacy_bin = bindir_p() / 'node_exporter'
+    if NODE_EXPORTER_BIN.exists() or (legacy_bin.is_file() and not legacy_bin.is_symlink()) or (bindir_p() / 'prometheus-node_exporter').exists():
```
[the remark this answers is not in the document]
defbin: [node_exporter_install: restore guard, check symlink](https://github.com/scylladb/scylla-monitoring/pull/2904/commits/972f530b8502f5de77e6b9f17d835cfd9c18140e)

packer/files/node_exporter_install
```diff
@@ -97,6 +99,8 @@ if __name__ == '__main__':
         run('install -d {dir}'.format(dir=NODE_EXPORTER_DIR))
         run('install -m 755 {src} {dst}'.format(
             src=os.path.join(extract_dir, NODE_EXPORTER_BIN_IN_DEB), dst=NODE_EXPORTER_BIN))
+        run('ln -sfn {bin} {link}'.format(bin=NODE_EXPORTER_BIN, link=bindir_p() / 'node_exporter'))
```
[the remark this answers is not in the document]
defbin: updated the PR description

defbin:
> @defbin did you test it?

@amnonh yes, tested on amd64 and arm64 without a packer build: ran `packer/files/node_exporter_install` as-is, with the real deb, in an Ubuntu 24.04 systemd container (same base as the image).

Steps:

1. Build a systemd image with the pip deps `scylla_monitoring_install_ami` installs:
   ```dockerfile
   FROM ubuntu:24.04
   RUN apt-get update && apt-get install -y systemd systemd-sysv python3-pip curl && pip3 install --break-system-packages psutil pyyaml distro traceback-with-variables
   CMD ["/sbin/init"]
   ```
   `docker build [--platform linux/amd64] -t ne-test .`
2. Start it and copy the scripts in:
   ```
   docker run -d --name ne --privileged --cgroupns=host -v /sys/fs/cgroup:/sys/fs/cgroup:rw ne-test
   docker cp packer/files ne:/opt/files
   ```
3. Run the checks with `docker exec ne ...`:

| Check | master | this PR |
|---|---|---|
| `/opt/files/node_exporter_install --arch <arch>`: service active, `:9100/metrics` up | pass | pass |
| `/usr/bin/node_exporter` points to `/opt/node_exporter/node_exporter` | missing | pass |
| Re-run without `--force` refuses; `--force` reinstalls | pass | pass |
| Scylla Cloud drop-in `ExecStart=/usr/bin/node_exporter --web.listen-address=127.0.0.1:9100` starts | **fails** | pass |
| `/usr/bin/node_exporter` symlinked to another binary, no `/opt` install: run without `--force` refuses | pass | pass |

The second commit keeps the original "already installed" guard. Excluding symlinks from it, as the first commit did, let a run without `--force` overwrite a foreign `/usr/bin/node_exporter` link (the Copilot finding).

packer/files/node_exporter_install
```diff
@@ -97,6 +99,8 @@ if __name__ == '__main__':
...
+        run('{link} --version'.format(link=bindir_p() / 'node_exporter'))
```
yaronkaikov: The PR description says a broken link "fails the image build", but I don't think it does.

`scylla_monitoring_install_ami.install_node_exporter()` (packer/files/scylla_monitoring_install_ami:170-173) wraps `run('{_HOME}/node_exporter_install')` in a `try/except Exception` that only prints the error. So if `/usr/bin/node_exporter --version` fails here, `check_call` raises, this installer exits non-zero, and the AMI build swallows it and carries on — publishing an image with no working node_exporter and a green packer run.

If this smoke test is meant to gate the build, that `except` block needs to re-raise (or the provisioner should call the script directly).

yaronkaikov: Ordering nit, related to the above: this check runs *before* the cleanup lines (`rm -rf {extract_dir}` and `os.remove(deb_file)`).

A failure here aborts with an uncaught `CalledProcessError`, leaving `/var/tmp/scylla-node-exporter-extract/` and the ~10MB `.deb` behind — and since the AMI provisioner swallows the failure, that garbage gets baked into the published image.

Moving the check after cleanup, or wrapping cleanup in a `finally`, avoids it.

packer/files/node_exporter_install
yaronkaikov, line 102: The link is created unconditionally and nothing ever removes it, which may not do what you want in the SCT flow described in the PR body.

When Scylla + Manager are installed on top of the monitoring image, the real `scylla-node-exporter` deb puts its binary at `/opt/scylladb/node_exporter/node_exporter`. This compat link keeps pointing at the monitoring image's private copy in `/opt/node_exporter`, so `/usr/bin/node_exporter --version` reports the monitoring build rather than the package's, and the link dangles if `/opt/node_exporter` is ever cleaned.

Pointing the link at the deb path when it exists, or documenting that the link is monitoring-image-only, would be safer.

Separately, worth knowing before closing MONITOR-95: the concrete hardcoded consumer is SCT's `set_web_listen_address` (sdcm/cluster.py), which sed-matches `ExecStart=/usr/bin/node_exporter  --collector.interrupts` in the unit file. The shipped unit now has `ExecStart=/opt/node_exporter/node_exporter ... --no-collector.interrupts`, so that consumer stays broken regardless of this symlink — the breakage is in the unit's `ExecStart`, which this PR doesn't touch.

packer/files/node_exporter_install
defbin: Not swallowed: [`install_node_exporter()`](https://github.com/scylladb/scylla-monitoring/blob/1864222a54f089a5de42f38b0f9b2887594d6979/packer/files/scylla_monitoring_install_ami#L199-L204) has had no try/except since [d95f6560](https://github.com/scylladb/scylla-monitoring/commit/d95f6560bfaa33acaa5db1f97bc2fe323d9f3910), and [`run()` re-raises](https://github.com/scylladb/scylla-monitoring/blob/1864222a54f089a5de42f38b0f9b2887594d6979/packer/files/scylla_monitoring_install_ami#L76). The [template](https://github.com/scylladb/scylla-monitoring/blob/1864222a54f089a5de42f38b0f9b2887594d6979/packer/scylla-monitor-template.json#L122-L139) keeps Packer's [default](https://developer.hashicorp.com/packer/docs/provisioners/shell) `/bin/sh -e`, so the build fails. Checked by making `/opt/node_exporter` noexec and calling `install_node_exporter()` from a `/bin/sh -e` script: `--version` raises `PermissionError` and the script exits 1.

defbin: Fixed: the [`--version` check](https://github.com/scylladb/scylla-monitoring/pull/2904/files#diff-e15cd350fd46e6c202859a2ee93c53098c06ee5f953b51d631dd36c75c2ee68fR103-R105) now runs after cleanup, so a failed check leaves no extract dir or deb behind.

packer/files/node_exporter_install
defbin: Intended: before 4.16, `/usr/bin/node_exporter` was the image's own binary too ([4.15.2](https://github.com/scylladb/scylla-monitoring/blob/4e1483dd65330ce9e9b9b9a7e14d5caddba74a90/packer/files/node_exporter_install#L66)), and the deb ships nothing under `/usr/bin`. The link is created at image build time, before any Scylla install, so a check for the deb path would always be false there. If `/opt/node_exporter` were removed, the [unit](https://github.com/scylladb/scylla-monitoring/blob/1864222a54f089a5de42f38b0f9b2887594d6979/packer/files/node-exporter.service#L8) would break too, so the link adds no new failure.

SCT's [`set_web_listen_address`](https://github.com/scylladb/scylla-cluster-tests/blob/8f0a762130e6d723e36dde075968da83bef8cceb/sdcm/cluster.py#L2097-L2107) runs only on DB nodes ([caller](https://github.com/scylladb/scylla-cluster-tests/blob/8f0a762130e6d723e36dde075968da83bef8cceb/sdcm/cluster_aws.py#L1188-L1191)). Its pattern matches none of the [4.15.2](https://github.com/scylladb/scylla-monitoring/blob/4e1483dd65330ce9e9b9b9a7e14d5caddba74a90/packer/files/node-exporter.service#L5), [4.16.0](https://github.com/scylladb/scylla-monitoring/blob/fac61b89b229a69d2af2d88b7e09a6316c3924e4/packer/files/node-exporter.service#L8) or [PR](https://github.com/scylladb/scylla-monitoring/blob/1864222a54f089a5de42f38b0f9b2887594d6979/packer/files/node-exporter.service#L8) units, so 4.16 didn't break it.

yaronkaikov (approved):

